PR #138, opened 2015-07-13
Changed let to var to increase browser compatibility.  Only the lates…

codecounselor:
…t versions of Chrome support this ES6 feature.  (https://kangax.github.io/compat-table/es6/)

I initially tried to follow the setup instructions, which were quite good, and run the app in firefox.  So I was a bit flummoxed when it didn't give me a login page.  After some debugging and searching I found this to be the problem.

### DIFF
--- a/app/components/nav-menu.js
+++ b/app/components/nav-menu.js
@@ -13,7 +13,7 @@ export default Ember.Component.extend(UserSession, {
     isShowing: false,
     
     _setup: function() {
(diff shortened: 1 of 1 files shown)


tangollama:
@codecounselor, thanks for the contribution and for taking the time to setup the project. I'm really looking forward working more closely in the weeks and months ahead. Cheers!

jglovier:
sean-connery-fist-pump-us-open

PR #152, opened 2015-07-27
YUIDocs and unit test cases

codecounselor:
I was playing around with YUI Docs which seem to work assuming you use the tags correctly (http://yui.github.io/yuidoc/syntax/index.html#primary-tags).

Here is a screenshot:
screenshot 2015-07-27 00 25 55
If you want to generate them, read this page: https://github.com/cibernox/ember-cli-yuidoc

Just wanted to share this, not sure it needs merged just yet.  But if you like it feel free.

Still trying to figure out the best stack for tests.  I don't know that I really like QUnit all that much.


jkleinsc:
@codecounselor Have you read the Ember.js testing guide:
http://guides.emberjs.com/v1.13.0/testing/

QUnit is the default testing framework used by Ember, though I would be open to using something else as long as you could still test via ember test.

codecounselor:
Yeah, how do you think I figured it out :) What I have to determine is if the helpers for the alternative frameworks (probably jasmine or mocha) are on par.  https://github.com/subeeshb/ember-jasmine 

I'm leaning towards Jasmine since I like the nested support for suites and setup/teardown and the general slant towards BDD.  Mocha/Chai/Sinon seems to also be a popular choice.

QUnit is older and entirely flat.  I'm also not sure what it provides in the way of mocks/fakes

tangollama:
@jkleinsc, let's confirm on this tomorrow. Any reason we shouldn't merge this PR?

taras:
I would argue to stay with QUnit to keep the project accessible to most Ember developers.

jkleinsc:
Closing this PR in favor of the work that @taras did.

PR #203, opened 2015-11-09
Medication module tests

zinyando:
This PR seeks to add tests for the medication module. Tests that will be added are

- [x] visiting /medication
- [x] creating a new medication request
- [x] fulfilling a medication request
- [ ] returning medication


jkleinsc:
I found the issue with the medication return and fixed it by changing the inventory-typeahead component to add a class when it has inventory items available.  Thanks @zinyando for the PR!

PR #222, opened 2015-12-06
Prevent all robots from scanning the site

mitchellcash:
- Add a comment making it clear to everyone that the directives from the robots.txt file allow all content on the site to be crawled

### DIFF
--- a/prod/robots.txt
+++ b/prod/robots.txt
@@ -1,3 +1,5 @@
 # http://www.robotstxt.org
+
+# Disallow crawling of all content
 User-agent: *
-Disallow:
+Disallow: /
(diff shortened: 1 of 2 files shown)


jkleinsc:
@mitchellcash thanks for this PR.  You have brought attention to something I wasn't thinking about.  HospitalRun sites really shouldn't be crawlable by robots because the content isn't intended for public consumption.  Would you be willing to change this PR to implement a robots.txt that denies all robots?

mitchellcash:
I was thinking that but being new to the project I was still trying to learn my way around. I will implement a change like you mention above.

mitchellcash:
@jkleinsc please see the updated commit

jkleinsc:
Looks good.  Thanks @mitchellcash!

PR #223, opened 2015-12-06
Refactor index.html

mitchellcash:
- Move `meta` information above `title` in index.html `head`
- Update initial-scale from 1 to 1.0 for accuracy but does not change behaviour

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -3,9 +3,10 @@
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <title>Hospital Run</title>
     <meta name="description" content="">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Hospital Run</title>
+    
     <link rel="icon" type="image/x-icon" href="/favicon.ico" />
     <link rel="shortcut icon" type="image/x-icon" href="/favicon.ico" />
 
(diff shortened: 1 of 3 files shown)


jglovier:
Hey @mitchellcash, thanks for the suggestions, but I'm not sure I understand the purpose of these changes?

I've been using the initial-scale meta for years and never have had a problem with simply a value of `1`, nor have [ever seen](https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag) [it recommended](https://css-tricks.com/snippets/html/responsive-meta-tag/) [to use as](https://developers.google.com/web/fundamentals/design-and-ui/responsive/fundamentals/set-the-viewport?hl=en) anything other than `1`, so I'm :-1: on that change.

As for moving `meta` above `title` I'm also :-1: on that one unless you have a great reason for. Can you explain what you are hoping to accomplish with that change?

mitchellcash:
Not a problem @jglovier I only recently came across this project and love the initiative and idea.

This is essentially just a refactor with no real performance gains apart from my opinionated view on the readability and use of the code.

I have come across those links you quote before and interestingly this refactor including the initial-scale meta change are based on another project from Google called the Web Starter Kit in which they highlight their opinionated best practices for building a great experience across many devices and performance oriented.

I am just being nit picky with this pull request though so I am happy for it to be closed if you feel it should be.

jkleinsc:
@mitchellcash thanks for your PR.  I am going to close this as index.html as it exists was generated via Ember CLI and follows the conventions suggested/generated by Ember.js

Issue #209, opened 2015-11-24
Alert given emails addresses when items are in danger of stocking out


tangollama:
Along with changes to #207, fire an alert to an email address when a stock item reaches less than 14 days of stock remaining. Along with this, we need a configurable way to both define who is alerted, the # of days as a threshold, and the A,B,C category of the item(s) to trigger the alert.

And yes, I know this leads into a larger alerting function in the system.

turboMaCk:
I think it makes sense to implement logic for `in danger of stocking out` on backed/couchdb side. It should be send as another property on model (`Boolean`?) than the component `inventory/rank-select` from my PR or any other component (adding icon `!` to name?) should display some indication directly in list/detail. This should be the first step before implementation of alerts and emailing. Let's close ranking thing first and then I might have a look if I can help with this one too.

turboMaCk:
:rocket: Lets start. Let me know if anyone has some ideas on this one.

turboMaCk:
First part done in: https://github.com/HospitalRun/hospitalrun-frontend/pull/232

tangollama:
Closing this issue. The remainder of the issues needs to be addressed by https://github.com/HospitalRun/hospitalrun-server/issues/6 and https://github.com/HospitalRun/hospitalrun-server/issues/18

PR #225, opened 2015-12-06
inventory items rank attribut

turboMaCk:
Implement rank (enum `['A', 'B', 'C']`) on inventory items.


app/components/inventory/rank-select.js
```diff
@@ -0,0 +1,16 @@
+import Ember from 'ember';
+import SelectValues from 'hospitalrun/utils/select-values';
```
turboMaCk: I've used this util to generate options collection.

turboMaCk:
I see you are [using mixins](https://github.com/HospitalRun/hospitalrun-frontend/blob/master/app/mixins/inventory-type-list.js) for defining available options / enums for select. My implementation is using wrapper component. I chose this because I want to do it in more component-centric way since there is no need to allow controllers access this options. I'm opened to discussion on this decision.

app/components/inventory/rank-select.js
```diff
@@ -0,0 +1,16 @@
...
+
+const c = Ember.computed;
+
+export default Ember.Component.extend({
+  label: 'Rank',
+  rankOptions: ['A', 'B', 'C'],
```
stefanpenner: lets put this in init or a CP, putting complex values on the prototype is likely to cause grief.

app/components/inventory/rank-select.js
```diff
@@ -0,0 +1,16 @@
+import Ember from 'ember';
+import SelectValues from 'hospitalrun/utils/select-values';
+
+const c = Ember.computed;
```
turboMaCk: I do not like use prototype extensions (`.property()`) anymore since they are globally defined and may be deprecated in the future. I like to define dependencies using `const` reference on top of file. Let me know if you're OK with this or it's against your coding style.

app/components/inventory/rank-select.js
turboMaCk: actual options (Array of Objects) is defined via CP in [here](https://github.com/HospitalRun/hospitalrun-frontend/pull/225/files#diff-0efd4da99b10af26a375b4c06eabe7baR13). I can define setter for this in `init()` if you like to.

turboMaCk:
### Design

This is how it looks like in lists:
screen shot 2015-12-06 at 17 15 34

this is edit form:
screen shot 2015-12-06 at 17 15 51

### Ideas on improvements:
We can add binding on `item.quantity` to `inventory/rank-display` component and compute classes (danger/success...) based on quantity and rank. But I'm nut sure if this should also depend on `Distribution Unit` type, so I'm leaving it as it is since this needs definition.

Let me know what you think about this PR. I'm ready to work on improvements based on your feedback.

tangollama:
@turboMaCk this screenshots look great. I'll review w/ @jkleinsc tomorrow and either accept the PR or get you feedback. Thanks for hopping right in.

app/components/inventory/rank-select.js
stefanpenner: a common idiom is to use destructuring.
```
const { computed } = Ember;
```

app/components/inventory/rank-select.js
```diff
@@ -0,0 +1,22 @@
...
+
+export default Ember.Component.extend({
+  label: 'Rank',
+  rankOptions: [],
```
stefanpenner: this looks like instance state, it should likely be instantiated in the initializer, or as a CP.

Placing instance state of the prototype can often lead to unexpected behavior and leaks)

app/components/inventory/rank-select.js
```diff
@@ -0,0 +1,22 @@
...
+  prompt: ' ',
+  class: 'col-sm-2 test-inv-rank',
+
+  options: c('rankOptions', function() {
+    return SelectValues.selectValues(this.get('rankOptions'));
+  }),
+
+  init() {
+    this._super();
```
stefanpenner:
```
this._super(...arguments)
```

app/components/inventory/rank-select.js
turboMaCk: ok

app/components/inventory/rank-select.js
turboMaCk: sure!

app/components/inventory/rank-select.js
stefanpenner: that being said, if this project is using the latest shims

https://github.com/ember-cli/ember-cli-shims/blob/master/app-shims.js#L194-L214

one can:

```js
import computed from 'ember-computed'
```

app/components/inventory/rank-select.js
turboMaCk: Imho setting this in initializer is little bit over-kill and made code confusing. Setting this to undefined/null by default do not specify type (maybe it will be better to remove this and keep only setter in init). The only reason I kept this is to let everyone know that this property will be and it should be `Array`. Computed property seems to be best solution but I'm not sure if there is no any reasonable overhead (since this type implements little bit more than we actually need). ES7 solution seems to be best for me, but I'm not sure, if it's compatible.

I can remove this completely (but for me it makes sense to have defined all props that instance has on one place) and keep the setter in `init()` or make CP (if it is not too much expensive)

I do not know as much about `ember-metal` as you so let me know your suggestions on this.

app/components/inventory/rank-select.js
turboMaCk: :+1:

app/components/inventory/rank-select.js
turboMaCk: BTW... this discussion is still actual even github hides this.

jglovier:
@turboMaCk thanks for your work here. Some design feedback:

image

These labels should use colors from the [existing color palette](https://github.com/HospitalRun/hospitalrun-frontend/blob/master/app/styles/_variables_mixins.scss) here. Although I looked through the code and it appears you're adding classes, but not defining colors anywhere? Now I'm wondering where those colors are coming from...

As for the edit screen, the only thing that concerns me is the Purchases table heading looks really crammed. It would be great to either reduce the label copy a bit (i.e. drop words where appropriate, like "Date" when date is obvious from the value), or find a way to better accommodate everything without it having to wrap to two lines of text everywhere.

jkleinsc:
@jglovier Those colors I believe are coming from bootstrap.  Also, the purchases table heading shown here has nothing to do with this PR.

@turboMaCk I am thinking we shouldn't display the ranking on the item listing.  I like what you've done there, but it's information that really is only needed for reporting and/or alerting purposes, so I don't think it belongs on the listing screen.

jglovier:
>Those colors I believe are coming from bootstrap

Ahhh, right.

>Also, the purchases table heading shown here has nothing to do with this PR.

Okay, cool. I'll address that separately then.

turboMaCk:
@jglovier is completely right. I used [bootstrap's labels](http://getbootstrap.com/components/#available-variations) for this. The only thing I had to change in form is `name` input and that `rank` select. I made a `name` input smaller.

We have to decide if we want to show that rank column in list tables. It also could be done some other way like I mentioned before it at least makes sense to me to add some `running out of` icon in the future.

I also want to continue and work on this https://github.com/HospitalRun/hospitalrun-frontend/issues/209 you after we finish this if want to.

turboMaCk:
If I should say my opinion on displaying this thing in list I will go this way:
 * keep displaying this information colorised in list since alter system is not done
 * after alert system is ready change this to displaying something like condition` based on rank and quantity.

jkleinsc:
@tangollama Any thoughts on this?

turboMaCk:
I will probably find some time to continue on this tomorrow. Let me know about changes :rocket:

jkleinsc:
@tangollama ping
I am of the opinion we do not include the rank on the inventory listing screen, primarily because there are a decent number of fields being displayed currently.  Also, if we are looking for a visual indicator, that could be applied to the quantity column, but I think that should be handled as a separate issue/PR

turboMaCk:
@jkleinsc so for now that setting will just do nothing. Ok, I do not thing that this is big deal so there is no need to sped much more time on this decision. I'll update this PR. and then ping for code review.

turboMaCk:
READY.

I also move component files to pod structure. I do not know why `ember g component` do not generate this by default.

turboMaCk:
rebased

app/inventory/rank-select/component.js
```diff
@@ -0,0 +1,21 @@
...
+
+  options: computed('rankOptions', function() {
+    return SelectValues.selectValues(this.get('rankOptions'));
+  }),
+
+  init() {
+    this._super(...arguments);
+
+    // set available options
+    this.set('rankOptions', Ember.A(['A', 'B', 'C']));
```
turboMaCk: because of this options can not be overwrite. Also there is unfinished discussion with @stefanpenner 

https://github.com/HospitalRun/hospitalrun-frontend/pull/225#discussion-diff-46771310
https://github.com/HospitalRun/hospitalrun-frontend/pull/225#discussion_r46775205

turboMaCk:
```
not ok 21 PhantomJS - Browser "phantomjs /home/travis/build/HospitalRun/hospitalrun-frontend/node_modules/ember-cli/node_modules/testem/assets/phantom.js http://localhost:7357/5350" exited unexpectedly.
```

jkleinsc:
@turboMaCk  Not sure why PhantomJS died -- maybe just a problem in Travis.  I re-ran the job and it looks good now.  As far as the options are concerned, one pattern we have used is to use mixins for hardcoded options.  If we ever want to change the options to pull from the database instead, we just change the mixin.

turboMaCk:
@jkleinsc thank you!

I know about mixins https://github.com/HospitalRun/hospitalrun-frontend/pull/225#issuecomment-162322175 I just waited for your feedback. Personally I like using components more than mixins in controllers (controllers are singletons & it's better isolated from other stuff) I can move this from component to mixin and use that mixin in that component but then there was no much other logic in component itself:D

Let me know what you think. I'll do this your way.

jkleinsc:
@turboMaCk I missed the comment about mixins earlier.  I think for now we should just leave it the way you coded it.  I will merge the PR unless you think there is another reason to wait.

turboMaCk:
I do not think that there are any blocker. I'll look at this after merge: https://github.com/HospitalRun/hospitalrun-frontend/issues/209

jkleinsc:
Looks good @turboMaCk.  Thanks for the PR!

turboMaCk:
My pleasure to helping you guys with this amazing project.

PR #229, opened 2015-12-07
Added Analytics

cd2-public:
Initial addition of Analytics via google to app

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -32,6 +32,14 @@
     <script>
         window.addEventListener('load',function(e){window.applicationCache.addEventListener('updateready', function(e){if (window.applicationCache.status==window.applicationCache.UPDATEREADY){window.applicationCache.swapCache();window.location.reload();}},false);},false);
     </script>
+    <script>
+      (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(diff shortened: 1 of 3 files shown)


cd2-public:
probably... I can iterate on this and come back, tbh I get the same error, but realtime analytics shows my visits pinging to google analytics dashboard, so it works as-is, it's just ugly.

jglovier:
I think we probably want to use [Event Tracking](https://developers.google.com/analytics/devguides/collection/analyticsjs/events) to get insights into where users are clicking in the UI.

Lewiscowles1986:
already mentioned this, all clicks are registered by google analytics automatically

turboMaCk:
At first you include analytics only in index.html not to tests/index.html. Also you have to define global var if you want to use this https://github.com/HospitalRun/hospitalrun-frontend/pull/229/files#diff-f3a289058604b2b069d07bb8e2cda60cR9

After all I do not think this is a good implementation.

1) What about using segment https://segment.com/?
2) I will rather define service for analytics stuff (something as a wrapper)

Lewiscowles1986:
going to be integrating ember-metrics package

turboMaCk:
looks like better solution

jkleinsc:
I am closing this PR in favor of using ember-metrics.

PR #232, opened 2015-12-10
Inventory items condition

turboMaCk:
related: https://github.com/HospitalRun/hospitalrun-frontend/issues/209

work in progress


turboMaCk:
nothing important here yet. I'm still looking around so I'm doing syntax refactoring at least..

turboMaCk:
Early prototype of model's `condition` attribute.

I have no clue how this should be set so for now only hardcoded intervals are used. It does not depend on `Distribution Unit` only on `quantity` and `rank`.

| x |  good |  average  |  bad |
|:-:|:-----:|:---------:|:----:|
| A |  >=20 | 100 - 199 | <100 |
| B | >=100 |  50 - 99  |  <50 |
| C |  >=50 |  25 - 49  |  <49 |

Maybe we should add new attr on inventory model and let user sets properties of this on each item (html rank select?)... but than rank enum (dropdown) will be useless.

app/models/inventory.js
```diff
@@ -1,7 +1,21 @@
...
+  quantity /= multiplier;
+
+  if (quantity >= 100) {
+    return 'good';
+  } else if (quantity < 50) {
+    return 'bad';
+  } else {
+    return 'average';
+  }
+}
```
turboMaCk: related code

app/models/inventory.js
```diff
@@ -28,26 +43,41 @@ export default AbstractModel.extend(LocationName, {
...
+      case 'A':
+        return getCondition(quantity, 2);
+      case 'B':
+        return getCondition(quantity);
+      case 'C':
+        return getCondition(quantity, 0.5);
+      default:
+        return;
+    }
+  }),
```
turboMaCk: related code

tangollama:
@turboMaCk, I appreciate the fact that you tried to sort this out w/o the advantage of good requirements work from me, and I'm sorry that was the case as well. Looking at this, I'm thinking that the issue of the stock position or "condition" of an inventory item is a little more nuanced. 

We have a report that calculates the number of days left of stock in a given item that uses past usage to predict a future stock out. It's called the Days Supply Left In Stock. 
screen shot 2015-12-11 at 1 31 39 pm
I'm thinking that the best way to judge a certain item's stock "condition" is gauge the number of days left in stock combined with the A,B,C rank. 

So ideally, the quantity would be replaced by the calculated number of days left in stock (something like estimatedDaysOfStock). So to do this properly, we would ideally like that to be the const on line 68 of app/models/inventory.js. Since that field doesn't exist yet, I think we can merge this PR, but them I'd like to propose that (if you're willing) you take on this issue: https://github.com/HospitalRun/hospitalrun-server/issues/6

Does that make sense?

turboMaCk:
Dont worry. This is a naive prototype. I just open PR to start discussion on this. This one will take a longer in my opinion. There will be a lot of tweaking I think.

turboMaCk:
@tangollama So it's done. There is temporary hardcoded value for `estimatedDaysOfStock` https://github.com/HospitalRun/hospitalrun-frontend/pull/232/files#diff-5a90295e0bf226b91b8c48e9d1e727e9R67

Let me know about settings (days for each rank to indicate each condition)

It's already rebased to current master.

app/models/inventory.js
```diff
@@ -28,26 +32,36 @@ export default AbstractModel.extend(LocationName, {
...
     return returnLocations.toString();
-  }.property('availableLocations'),
+  }),
+
+  condition: computed('rank', 'estimatedDaysOfStock', function() {
+    const estimatedDaysOfStock = this.get('estimatedDaysOfStock');
+    const multiplier = rankToMultiplier(this.get('rank'));
+
+    return getCondition(estimatedDaysOfStock, multiplier);
+  }),
```
turboMaCk: implementation on model

app/utils/item-condition.js
```diff
@@ -0,0 +1,33 @@
...
+  estimatedDaysOfStock *= multiplier;
+
+  if (estimatedDaysOfStock >= 14) {
+    return 'good';
+  } else if (estimatedDaysOfStock < 7) {
+    return 'bad';
+  } else {
+    return 'average';
+  }
+}
```
turboMaCk: utils for calculations on model.

turboMaCk:
rebased again to @jkleinsc's https://github.com/HospitalRun/hospitalrun-frontend/commit/77256ae018e6cf376496c42b83d8f4b576106379

jkleinsc:
@turboMaCk Wow that was fast.  I'm taking a look now at merging this PR in.

turboMaCk:
I was working on it at the time you push new commits to master. I'm online and ready to work on any improvements based on your decision right now so let me know what you think. It will be nice to push this forward and move to [next step](https://github.com/HospitalRun/hospitalrun-server/issues/6) soon.

jkleinsc:
@turboMaCk I agree.  I am going to merge this and if we need to change settings, we can do that later.

jglovier:
:metal: :zap:

PR #265, opened 2016-01-18
Add Service Worker support

jkleinsc:
This PR adds service worker support to HospitalRun Frontend.

Right now this means that in browsers that support service workers (see https://jakearchibald.github.io/isserviceworkerready/), we are doing the following:
1. Offline caching of the application will be handled with a service worker instead of through application cache.  [Firefox](https://twitter.com/reactjs_news/status/685875497473974272) and [Chrome](https://twitter.com/jaffathecake/status/685595711401005057) are planning on deprecating application cache in favor of service workers.
2. Syncing data to/from pouchdb/couchdb now runs in the service worker.  This means that the sync process is running in a separate thread which should be more performant.
3. When the user is online, data is read/written directly to couchdb.  If the user is offline data is read/written from pouchdb.
4.  Service workers are HTTPS only but will work on http in development on localhost.  

I think we can merge this PR as is, but would like some feedback/signoff from @tangollama and @jglovier as well as anyone else who has thoughts on this.

Also, I am thinking that we should make the offline sync a configurable option so that implementations that would prefer to not store data offline can do so.


jglovier:
@jkleinsc hellll yeah offloading processes!! :metal: :tada: 

>Also, I am thinking that we should make the offline sync a configurable option so that implementations that would prefer to not store data offline can do so.

:+1: :+1: 

:shipit:

jonkoops:
@jkleinsc I'd suggest you remove the merge commits where you pulled in changes and rebase your earlier commit on top of the upstream master. This way the master branch history won't get polluted with unnecessary merge commits.

jkleinsc:
@jonkoops yeah... I'm planning on doing a git merge --squash when we merge this PR in.

jkleinsc:
@jglovier @tangollama what do you think about merging this PR in?

jglovier:
@jkleinsc :ship:

tangollama:
@jkleinsc make it so.

Issue #277, opened 2016-01-22
Localization for Imaging Module


jkleinsc:
Now that i18n support has been added to HospitalRun via PR #248 and issue #141 we need to localize the app.  This issue is to track localization for the imaging module.  Localization in HospitalRun uses [Ember-I18n](https://github.com/jamesarosen/ember-i18n).  The translations/default text will be placed in app/locales/en/translations.js.
See https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Translating-Text for information on how to localize the templates.

alexpelan:
I can take a stab at this one this weekend whilst snowed in - imaging looks like the simplest module, so easiest to get a feel for the i18n process.

PR #283, opened 2016-01-25
Address #277: Localization for Imaging Module

alexpelan:
-Added an initializer to inject the I18N service into all controllers, routes, and mixins
-Did not write any i18n-specific tests - don't have a lot of experience in how to automated test i18n, in this case knowing that the current tests didn't break when run with the 'en' locale is probably good enough? Would gladly write some if you have any suggestions.
-The Ember localization syntax can get kind of awkward sometimes, it doesn't feel very DRY. Take for example the subActions changes in app/imaging/route.js - you can't access the translation helper in the array because this isn't bound to the route , so you have to change it to a function. Not sure if there's a way around this, that method was recommended here: http://stackoverflow.com/a/33380796/265714

A couple of things I considered out of scope:
-I didn't localize the nav
-I made no effort to audit display of numbers / dates / etc in the proper locale - seems like we're just focused on translating strings for now.

### DIFF
--- a/app/components/nav-menu.js
+++ b/app/components/nav-menu.js
@@ -27,7 +27,6 @@ export default Ember.Component.extend(UserSession, {
 
   actions: {
     toggleContent: function() {
-      // debugger;
       this.set('isShowing', !this.get('isShowing'));
       this.sendAction('callNavAction', this.nav);
     },
--- a/app/controllers/abstract-edit-controller.js
+++ b/app/controllers/abstract-edit-controller.js
@@ -7,11 +7,12 @@ export default Ember.Controller.extend(EditPanelProps, IsUpdateDisabled, ModalHe
   cancelAction: 'allItems',
 
   cancelButtonText: function() {
+    let i18n = this.get('i18n');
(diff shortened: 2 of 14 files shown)


jkleinsc:
@alexpelan Good work here.  
As far as i18n specific tests, I think for now we are ok.  Eventually I would like to add tests to verify that translations aren't missing, but I think that is outside the scope of this issue.
Using a computed property (i.e. `function(){}.property()`) is the accepted way to tackle the subActions.
You are correct that the other stuff you listed is out of scope for this issue.

If you can clean up the tests that are failing (looks like some simple formatting issues) and remove the //debugger from app/components/nav-menu.js, I think we can merge this PR.

alexpelan:
@jkleinsc Just force pushed and it looks like everything passed. I think you read the diff wrong - i removed the //debugger, didn't add it - unless that was a load-bearing //debugger :-)

tangollama:
+1 XP to @alexpelan for witty PR comments.

jkleinsc:
@alexpelan PR looks good.  Yup... I misread that and yeah it wasn't a load-bearing //debugger!

Issue #44, opened 2015-04-02
Appointment edit from patient not returning to patient


jkleinsc:

capurp:
I would like to take a shot at fixing this issue and officially put my name on it.

tangollama:
Great!

Typed or dictated imperfectly on my iPhone (717.385.9970)

PR #292, opened 2016-01-30
Appointment edit from patient fix

capurp:
Address Issue #44 

I was able to reproduce the error by following the following steps:

1. create a new patient
2. navigate to appointment
3. click on new appointment
4.  fill in a new appointment for the patient just created (other parameters do not matter)
5. From the new appointment screen click ‘add’ 
6. After getting to appointment summary for this week navigate to patient list
7. find the newly created patient and click ‘edit’
8. Navigate to ‘Visits/Appointments’
9. click on ‘edit’ for the newly created appointment
10. from this screen click ‘return’ 

expected behavior: returns to edit patient screen

actual behavior: returns to the appointment summary screen

However, I noticed that this error did not occur all the time. If return to the patient listing and edit another patient’s appointment from the patient edit screen, the return button behaves as expected. 

After taking a look at the appointment/edit controller with ember inspector, I pinpointed the issue to the `cancelAction` Property not being set to `returnToPatient` when trying to edit the appointment immediately after creating the appointment. Upon further inspection I found that this property is a computed property that was not recalculated when following the steps outlined above. 

I did some research and I believe that this issue is similar to the issue outlined here:

http://balinterdi.com/2014/06/26/ember-gotcha-controllers-are-singletons.html

It seems that the controller cached the computed property when you navigate to the create appointment and does not recognize that one of the dependent properties has changed. 

The action `editAppointment` sets the property `returnToPatient` to true before transitioning to appointment.edit route. However this does not trigger the computed property `cancelAction` to recalculate. My solution is to notify the controller that the property changed manually in the setupController hook. This is successful in refreshing the `cancelAction` property with the updated model property.  However I do not know if this is the appropriate place or the best way to accomplish this fix. Any feedback is greatly appreciate. Thank you


jkleinsc:
@capurp thanks for the PR.  While this PR may fix the issue, my concern is that making the change at the abstract-edit-route.js affects other routes that extend the abstract-edit-route.  

Also, looking at the cancelAction computed property on the appointment edit controller, it looks like the real issue is that if a property called `returnTo` is set on the model, that value is used.  When you view an appointment from one of the appointment listing screens (eg Appointments This Week), it sets this value.  The problem is that the value doesn't get reset when viewing the appointment from the patient's screen.

I believe the proper fix would be to clear out the `returnTo` property from the appointment model in the `editAppointment` action in the patient edit controller (/app/patients/edit/controller.js).

Does that make sense?

capurp:
@jkleinsc thanks for your help with another issue and thank you for your feedback! I applied the change to clear out the `returnTo` property in the `editAppointment` action and huzzah! It works!

I see how the solution makes more sense after taking a look at the edit appointment controller. This is slightly off topic, but now I am slightly confused as how my first proposed fix worked at all in the first place! If the edit appointment controller is a subclass of the patient-submodule because it uses the `extend()` method, wouldn't notify property change on `cancelAction`  invoke the `cancelAction` on the edit appointment controller resulting in the same `cancelAction`?

jkleinsc:
@capurp thanks for the update to the PR.  I'm not sure how your first fix worked and I am wondering if it didn't fix the problem, but based on the testing route you took, it may have looked like it was fixed.
The problem only presents itself if you view an appointment from the appointment listing screen (eg Today's Appointments) and then view that same appointment from the patient screen.  

I'm going to go ahead and merge this PR.

PR #295, opened 2016-02-05
A way to better manage dependencies

filipesperandio:
Simplify DEV env setup a bit
- `make serve` starts couchdb & server, resolve dependencies like bower and ember-cli too - requires `couchdb` in the path (if installed through `apt-get` or `brew` you are all done)
- `make all` run all tests and build the app

Tested on Mac OS 10.11.13 and Ubuntu 14.04


README.md
```diff
@@ -13,23 +13,21 @@ To install the frontend please do the following:
 - Make sure you have installed [Git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
 - Make sure you have installed [Node.js](https://nodejs.org/en/download/). Versions after 0.10.0 should work, but please note if you encounter errors using 5.x it may be necessary to upgrade your npm version. Versions after 3.5.x should work:
     1. `npm install -g npm`
-- Install [ember-cli v1.13.14](https://www.npmjs.org/package/ember-cli): `npm install -g ember-cli@1.13.14`
```
filipesperandio: Does it have to be this specific version?

jkleinsc: @filipesperandio we could restate this as install the latest ember-cli since we are keeping it up to date as much as possible.

jkleinsc:
@filipesperandio is this PR an OS specific PR?  I tested it on OS X 10.10.5 and it doesn't seem to work there.  I am wondering if there is a better way to do this in a way that would work across multiple OSes.

filipesperandio:
@jkleinsc Should not depend on OS. Could you tell me what issues you face?

jkleinsc:
@filipesperandio I just tried `make serve` and `make couchdb` and both tell me I need to install couchdb which is already installed and running.  Looking at script/couchdb I think the problem there is that couchdb executable isn't in the path on a default install (eg if I try to run `couchdb` from the command line it doesn't run).

I'm also wondering if we could just do this all under npm vs make.  While I'm comfortable using make, I think there may be some folks who would be uncomfortable using it.

I am going to throw this issue into our slack channel as I would like to get more feedback from the community.

README.md
jonkoops: `npm install -g ember-cli@latest` should do the trick.

jkleinsc:
@filipesperandio I like the changes.  I am wondering if for couchdb check we can change it do the curl check that you put in for npm start instead of using command -v?  That way it doesn't matter if the binary is located in the path.  Also, I'm not that familiar with how to build makefiles so I may be misunderstanding what it is doing, but ember-cli, bower and phantomjs should be installed globally via npm install -g.  Maybe there isn't a way to make that work with make?  
Also, I am curious about the bower bit at the bottom:
```
$(bower): node_modules bower.json
	@bower install --allow-root
	@touch $@
```
Why does bower have reference to node_modules?

filipesperandio:
@jkleinsc The `command -v` checks if `couchdb` executable is in the path so the script can proceed and call it, checking if the service is running won't guarantee that.

Sorry if this PR is introducing too much in terms of changes in the DEV toolset, I hope it is for the best.

Letting `make` manage the dependencies also means no global dependencies, so Make will rely on `bower`, `phantomjs` and `ember-cli` installed under the local `node_modules` directory.
The benefit is that we can include `node_modules/.bin` within Make's path so we always use versions we specified or that are under the project's control.

Make is a file generator, to simply put, their targets are, basically, files. As a file generator, the same file won't be generated more than once if things it depends on stay the same, for that it checks file's timestamps.
`$(bower)` is a target that refer back to the `bower` variable defined on the beginning of Makefile. It is the directory where bower modules are installed (pointed by `.bowerrc`).
The dependency on `node_modules` makes sure `npm install` have ran before, so `bower` binary is installed. `bower.json` is also a dependency, so if it gets changed, the command in the following line is executed next time `make` needs to fulfill a dependency. `touch` is only there because the target is a directory and the command is changing things within it, so the timestamps might match.
Back to `node_modules` dependency, it just runs once, or if its own dependency changes, in this case `package.json`.
For example, if you update `package.json` to use a different version of a lib, do the same for a bower dependency and simple run `make serve` after that, Make will figure `npm install` must run before `bower install` and just after that `./script/serve`, but if nothing changes, only the last will get executed.

filipesperandio:
@jkleinsc Glad you pointed it out, it was not 100% correct making `$(bower)` depend on `node_modules`. Since we don't refer `bower` on `package.json` I've changed the dependency to be `node_modules/bower` instead. The principle still the same, just the target that changes a bit.

I would advocate for keeping all the dependencies in the `package.json` but we can cycle this back later on.

jkleinsc:
@filipesperandio I am going to go ahead and merge this PR.  One question on your Mac OS testing -- how did you install couchdb and/or did you explicitly add couchdb to the path or did that happen automatically on install?

jkleinsc:
@filipesperandio I noticed something when I was merging this PR -- make serve doesn't do anything with server/config.js but the readme seems to indicate you can skip step 7, which is to:

> Copy the server/config-example.js to server/config.js.

So either we need to update the readme or change the makefile to copy server/config-example.js to server/config.js if it doesn't exist

filipesperandio:
I've installed couchdb through [homebrew](http://brew.sh/): `brew install couchdb` and the `path` was automatically setup.

Step 7 addressed in the Makefile, thanks for catching that.

filipesperandio:
I had updated the PR descriptions a couple days ago with a bit more detail around requirements and tests I did, just for the record.

PR #300, opened 2016-02-12
Added i18n error messages on login page

nerdkid93:
Instead of displaying error message from Pouch when login fails, it displays an error message in translations.js.


nerdkid93:
The error is in the test. Part of this commit is changing the error message when a user fails to login. The actual output should be the expected output.

jkleinsc:
@nerdkid93 Can you update the failing test as part of this PR?
The failing test is located here:
https://github.com/HospitalRun/hospitalrun-frontend/blob/master/tests/acceptance/login-test.js

nerdkid93:
Yup, I'll take care of that tonight.

nerdkid93:
Test has been updated, and all tests were passed.

jkleinsc:
@nerdkid93 looks good. I'll merge it in.  Thanks for the PR!

PR #316, opened 2016-02-25
Add russian localization

G07cha:
Feel free to ping me when localization base will expand.


jkleinsc:
Спасибо! Thanks! @G07cha for the PR

Issue #141, opened 2015-07-16
Add internationalization support


jkleinsc:
In order to support different languages in HospitalRun we need to implement [ember-i18n](https://github.com/jamesarosen/ember-i18n) throughout the app

billybonks:
Hi, i have been looking for some way to contribute to the project, and i think this is a good issue to work on, since it's a nice way to explore the code base :smile:

jkleinsc:
@billybonks One of the things on my todo list is to upgrade to Ember 1.13, then fix the deprecation warnings in preparation for Ember 2.x

Do you know how much of a difference there is between 3.x i18n and 4.x?  If they are significantly different, I would prefer to wait until we can pull in 4.x.

tangollama:
@billybonks, given that this conversation went cold back in July b/c of the upgrade work, I'm guessing you didn't start anything with this; but I wanted to ask if this was on your radar or if I should be looking for someone else to look into it? No worries either way. Just asking.

billybonks:
@tangollama, let me try set this up this weekend otherwise, ill drop it :+1: my time has been limited lately

tangollama:
Word.

jonkoops:
@billybonks Any update on this?, would love to help out.

jkleinsc:
@jonkoops @billybonks had a PR that setup localization and that has been merged and at this point what is needed is to convert all of the templates to use localized text.  I was intending on creating issues for each functional area to convert the templates but I haven't done so yet.

jonkoops:
@jkeinsc Alright, let me know when you do then I'll pick some up.

jkleinsc:
@jonkoops The following localization issues have been added:
#274, #275, #276, #277, #278, #279, #280, 281

mDantas:
@jkleinsc @jonkoops just wanted to say hi and that'd I'd be interested in helping out too.

jonkoops:
@mDantas I'll take the Patients module, maybe you can work on the Appointments module at the same time?

jkleinsc:
Once all the localization has been completed, we should add [ember-cli-template-lint] https://github.com/rwjblue/ember-cli-template-lint to verify that all of our templates are using translated strings.

jkleinsc:
All issues complete.  Closing this issue.

PR #317, opened 2016-02-27
i18n support localization for navigation

James1x0:
Was going to work on labs localization but realized most of the app is still unlocalized _top-down_. This pull addresses the side navigation specifically via the navigation mixin.

- Add support for ember-i18n localization on navigation items & sub-items
- Add localization support for about modal title
- Added EN locale for navigation & about title

Related to broad issue #141


jkleinsc:
@James1x0 looks good.  Thanks for the PR!  I'll merge it in.

PR #318, opened 2016-02-27
Fixes #278: Localization for Medication module

### LINKED ISSUE #278: Localization for Medication Module
Now that i18n support has been added to HospitalRun via PR #248 and issue #141 we need to localize the app.  This issue is to track localization for the medication module.  Localization in HospitalRun uses [Ember-I18n](https://github.com/jamesarosen/ember-i18n).  The translations/default text will be placed in app/locales/en/translations.js.
See https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Translating-Text for information on how to localize the templates.

alexpelan:
One thing I wanted to highlight is the 'searchStatus' parameter used in many of the index routes isn't easy to internationalize right now - because that same string is used for both display and is actually stored in the model - I think we'll have to switch to use category id: string mappings and translate those for the models?

### DIFF
--- a/app/locales/en/translations.js
+++ b/app/locales/en/translations.js
@@ -99,10 +99,13 @@ export default {
     patient: 'Patient',
     quantity: 'Quantity',
     requested_on: 'Requested On',
+    date: 'Date',
     date_requested: 'Date Requested',
     date_completed: 'Date Completed',
     requested_by: 'Requested By',
     fulfill: 'Fulfill',
+    fulfill_request: 'Fulfill Request',
+    fulfill_request_now: 'Fulfill Request Now',
     actions: 'Actions',
     action: 'Action',
     notes: 'Notes',
@@ -134,7 +137,16 @@ export default {
     password: 'Password',
     edit_user: 'Edit User',
     new_user: 'New User',
-    delete_user: 'Delete User'
+    delete_user: 'Delete User',
+    medication: 'Medication',
+    status: 'Status',
+    add_new_outpatient_visit: '--Add New Outpatient Visit--',
+    prescription: 'Prescription',
+    prescription_date: 'Prescription Date',
+    bill_to: 'Bill To',
+    pull_from: 'Pull From',
+    fulfilled: 'Fulfilled',
+    delete_request: 'Delete Request'
   },
   messages: {
     no_items_found: 'No items found.',
@@ -143,7 +155,11 @@ export default {
     no_users_found: 'No users found.',
     are_you_sure_delete: 'Are you sure you wish to delete the user ',
     user_has_been_saved: 'The user has been saved.',
-    user_saved: 'User Saved'
+    user_saved: 'User Saved',
+    new_patient_has_to_be_created: 'A new patient needs to be created...Please wait..'
+  },
+  alerts: {
+    please_wait: 'Please Wait'
   },
   buttons: {
     complete: 'Complete',
@@ -156,7 +172,10 @@ export default {
     new_user: 'New User',
     add_value: 'Add Value',
     import: 'Import',
-    load_file: 'Load File'
+    load_file: 'Load File',
+    new_request: 'New Request',
+    all_requests: 'All Requests',
+    dispense: 'Dispense'
   },
   login: {
     messages: {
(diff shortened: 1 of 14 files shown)


jkleinsc:
@alexpelan unless I am missing something the medication module doesn't display searchStatus anywhere.  The only place I see it being displayed is on the invoices index page.

alexpelan:
@jkleinsc Sorry if I was being unclear - 'searchStatus' appears (to me) to be used to filter which medication records show up - it looks like it searches for the model's 'status' property. The model's status property appears to be an english string, and is then displayed in the UI, for example in the medication/index template.  These localization tickets appear to specifically target the UI text, I was just pointing out there's some other work that will have to be done.

app/locales/en/translations.js
```diff
@@ -78,5 +93,47 @@ export default {
...
+      credit_to_account: 'Credit To Account'
+    },
+    alerts: {
+      returned_title: 'Medication Returned',
+      returned_message:  'The medication has been marked as returned.',
+      saved_title: 'Medication Request Saved',
+      saved_message: 'The medication record has been saved.',
+      fulfilled_title: 'Medication Request Fulfilled'
+    },
+    common: {
```
jkleinsc: @alexpelan was this common section intended to be labels common across the app?  It is nested under medication in this file.

app/medication/delete/controller.js
```diff
@@ -1,7 +1,8 @@
+import { translationMacro as t } from 'ember-i18n';
 import AbstractDeleteController from 'hospitalrun/controllers/abstract-delete-controller';
 import PatientSubmodule from 'hospitalrun/mixins/patient-submodule';
 export default AbstractDeleteController.extend(PatientSubmodule, {
-  title: 'Delete Request',
+  title: t('common.delete_request'),
```
jkleinsc: @alexpelan here common isn't nested.

app/locales/en/translations.js
jkleinsc: Also, if it is intended to be app wide, a label other than "common" would be better.

app/medication/edit/controller.js
```diff
@@ -69,22 +70,24 @@ export default AbstractEditController.extend(InventorySelection, FulfillRequest,
...
+      alertTitle = i18n.t('medication.alerts.saved_title');
+      alertMessage = i18n.t('medication.alerts.saved_message');
     }
     this.saveVisitIfNeeded(alertTitle, alertMessage);
   },
 
   _addNewPatient: function() {
-    this.displayAlert('Please Wait', 'A new patient needs to be created...Please wait..');
+    let i18n = this.get('i18n');
+    this.displayAlert(i18n.t('common.please_wait'), i18n.t('messages.new_patient_has_to_be_created'));
```
jkleinsc: @alexpelan again common isn't nested - I'm assuming your intention was to not nest it?

app/locales/en/translations.js
alexpelan: It wasn't intended to be nested - great catch, thanks. What would you suggest as a better label? I was looking for something that could be a button, or a label, or a title

jkleinsc: I think it would be better to place them in more specific buckets such as button label or title because it gives translators better context of the intention of that text.  Does that make sense?

jkleinsc:
@alexpelan gotcha about searchStatus.  Yes the problem is the text in the model.

app/medication/edit/route.js
```diff
@@ -1,12 +1,13 @@
...
 import Ember from 'ember';
 import FulfillRequest from 'hospitalrun/mixins/fulfill-request';
 import InventoryLocations from 'hospitalrun/mixins/inventory-locations'; // inventory-locations mixin is needed for fulfill-request mixin!
 import PatientListRoute from 'hospitalrun/mixins/patient-list-route';
 export default AbstractEditRoute.extend(FulfillRequest, InventoryLocations, PatientListRoute, {
-  editTitle: 'Edit Medication Request',
+  editTitle: t('medication.titles.edit_medication_request'),
   modelName: 'medication',
-  newTitle: 'New Medication Request',
+  newTitle: t('medications.titles.new_medication_request'),
```
jkleinsc: This should be  `t('medication.titles.new_medication_request'),`

alexpelan:
@jkleinsc responded to your feedback - thanks for the quick review.

jkleinsc:
@alexpelan looks good.  Thanks for the PR!

PR #319, opened 2016-02-29
Added Turkish localization

aeke:
Added Turkish localization file


jkleinsc:
Thanks for the PR @aeke !

PR #320, opened 2016-03-01
Localization for Admin Module

renato04:
address #[281]

This is the localization for the Admin Module.

### DIFF
--- a/app/users/delete/template.hbs
+++ b/app/users/delete/template.hbs
@@ -5,6 +5,6 @@
     updateButtonText=updateButtonText }}
   <div class="alert alert-danger">
     <span class="glyphicon glyphicon-warning-sign"></span>
-    Are you sure you wish to delete the user {{model.name}}?
+    {{t 'messages.are_you_sure_delete'}}{{model.name}}?
   </div>
 {{/modal-dialog}}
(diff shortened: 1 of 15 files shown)


jkleinsc:
Thanks for the PR @renato04.  I'll merge and resolve the conflicts.

app/users/delete/template.hbs
```diff
@@ -5,6 +5,6 @@
     updateButtonText=updateButtonText }}
   <div class="alert alert-danger">
     <span class="glyphicon glyphicon-warning-sign"></span>
-    Are you sure you wish to delete the user {{model.name}}?
+    {{t 'messages.are_you_sure_delete'}}{{model.name}}?
```
cfleschhut: @renato04 Quick notice: With ember-i18n we can interpolate values in translation strings (as to move the question mark into the string) like this: https://github.com/cfleschhut/hospitalrun-frontend/commit/0c6dbb9193739089e51e04ffc459b379ab883f3e :)

renato04: @cfleschhut  Thank you for the suggestion i didn't know that, i'll try to use this technique next time.

PR #327, opened 2016-03-03
Update README.md

akiaksdon:


jkleinsc:
The comment you have added 
>  (needs root access to install PhantomJs)

is not necessarily true, as it just depends how you have NPM setup.  See here for more details:
https://docs.npmjs.com/getting-started/fixing-npm-permissions

jkleinsc:
@akiaksdon I've given this some more thought and I am going to merge with a slight refinement:
Depending on your [npm permissions](https://docs.npmjs.com/getting-started/fixing-npm-permissions) you might need root access to install PhantomJS2.

PR #354, opened 2016-03-15
i8ln for the admin module

dswanson:
#345

### DIFF
--- a/app/admin/address/route.js
+++ b/app/admin/address/route.js
@@ -1,9 +1,10 @@
 import AbstractEditRoute from 'hospitalrun/routes/abstract-edit-route';
 import Ember from 'ember';
+import { translationMacro as t } from 'ember-i18n';
 export default AbstractEditRoute.extend({
   hideNewButton: true,
-  newTitle: 'Address Options',
-  editTitle: 'Address Options',
+  newTitle: t('admin.address.route.new_title'),
+  editTitle: t('admin.address.route.edit_title'),
   model: function() {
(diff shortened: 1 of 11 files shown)


billybonks:
So my issue with this implementation which is my fault, since i didn't add any documentation during the initial implementation.

initially i designed the localization's file to have two main categories.
-labels
-messages
-titles

since then people have added 
-buttons
-alerts

which is fine , they would have been merged into labels and messages.


any translation in these categories, if used in multiple places around the application would be in the first level of the hash. if they are only used in a specific place, they would be nested under there area for example a button in appointments would go under 

appointments -> buttons -> myButton.


in this pr you are nesting them under admin, you have added `route`, i can understand why.

you have also added loaddb, and lookup. which i am not so sure about. i feel most of these could go under messages -> errors or something like that.

i am just opening up conversation, so that we can standardize the file.

jkleinsc:
@billybonks we are having an ongoing conversation on how to best organize the localizations.  There are a couple of things we are thinking about:
1. Reuse can potentially be an issue because depending upon language/context you may not want to reuse the value.  I think there are some use cases where reuse is ok, but it probably should be used in moderation.
2. Dividing the localizations into separate files per module and then importing them into the main translations file.

If you have additional thoughts, please feel free to bring them up in slack.  I appreciate your feedback and input!

jkleinsc:
@dswanson I believe we should remove the .route and .controller namespaces

PR #372, opened 2016-03-17
Action cursor

pete-the-pete:
This adds a selector to highlight actionable areas by showing the pointer cursor for elements with the `data-ember-action` attribute.

This should probably be temporary, to help show where more specific classes could be added.  

Without the class
arrow

With the class
pointer

### DIFF
--- a/app/styles/_base.scss
+++ b/app/styles/_base.scss
@@ -4,4 +4,14 @@ body {
   &.ember-application {
     padding-left: 275px; // to accomodate for .sidebar-nav
   }
+
+  /*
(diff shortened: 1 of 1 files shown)


pete-the-pete:
I can/will handle the conflicts and merge the branch if accepted.

jglovier:
@pete-the-pete thanks! :tada:

jglovier:
I just realized using the attribute selector with `data-ember-action` results in lots of other elements besides the index rows matching the cursor declaration, including elements that aren't actually clickable. I've got a banch open for some cleanup, so I'll add a classname to the rows and target based on the class.

PR #402, opened 2016-04-04
Adding 'missed appointments' tab

### LINKED ISSUE #351: List of missed appointments
Create a new route in Appointments that lists the missed appointments. (suggested: /appointments/missed). The purpose for this is to create a mechanism for followup.

pete-the-pete:
Fixes #351 .

I didn't realize the other appointment routes already filter to the latest, so 'missed' doesn't require any filters.

I added a test, and modified the `createAppointment` helper function to accept params.

-

### DIFF
--- a/app/appointments/item/template.hbs
+++ b/app/appointments/item/template.hbs
@@ -6,7 +6,7 @@
     <td>{{appointment.appointmentType}}</td>
     <td>{{appointment.location}}</td>
     <td>{{appointment.provider}}</td>
-    <td>{{appointment.displayStatus}}</td>
+    <td class="appointment-status">{{appointment.displayStatus}}</td>
     <td>
         {{#if canAddVisit}}
             <button class="btn btn-default" {{action 'createVisit' appointment bubbles=false }}>{{t 'buttons.add_visit'}}</button>
--- a/app/appointments/missed/controller.js
+++ b/app/appointments/missed/controller.js
@@ -1,0 +1,4 @@
+import AppointmentIndexController from 'hospitalrun/appointments/index/controller';
+export default AppointmentIndexController.extend({
+  startKey: []
+});
--- a/app/appointments/missed/route.js
+++ b/app/appointments/missed/route.js
@@ -1,0 +1,17 @@
+import AppointmentIndexRoute from 'hospitalrun/appointments/index/route';
+import { translationMacro as t } from 'ember-i18n';
+
+export default AppointmentIndexRoute.extend({
+  editReturn: 'appointments.missed',
+  modelName: 'appointment',
+  pageTitle: t('appointments.missed'),
+
+  _modelQueryParams() {
+    let queryParams = this._super(...arguments);
+    queryParams.filterBy = [{
+      name: 'status',
+      value: 'Missed'
+    }];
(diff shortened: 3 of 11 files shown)


jkleinsc:
@pete-the-pete missed appointments doesn't mean all appointments in the past, it means appointments that were scheduled and the patient didn't show up.  The appointments missed should show all appointments that have a status of Missed, which will be a new status that will need to be added to hospitalrun/mixins/appointment-statuses.  You can then pass a filterBy value of Missed to the _modelQueryParams defined for the route, eg:
```javascript
      options: {
        startkey: [startOfWeek, null, null],
        endkey: [endOfWeek, endOfWeek, maxId],
        filterBy: [
          name: 'status',
          value: 'Missed'
        ]
      },
```

Also, it looks like the tests are failing.

pete-the-pete:
Ah...thanks for clarifying, and pointing me in the right direction.  I'll try to get something up soon.

pete-the-pete:
@jkleinsc, finally back from vacation....

I don't understand how an appointment can be aware if a patient showed up or not.  My thought was to see if a given appointment had any **visits**, but there isn't a relationships between the two.  Should there be?

I could add it, so that there is a hasMany relationship between **appointments** and **visits**.  Appointments that have no visits, and the the current data is past the appointment date would then be set to "Missed" (the status would be updated when visiting the appointments route).  Thoughts?

jkleinsc:
@pete-the-pete the key here is instead of looking at this as something that the system automatically determines, the marking of an appointment as missed is something that requires human interaction.  

I suppose at some point we could make this an automated process, but at this point using the status already used on an appointment is an appropriate way to handle this.
cc @HospitalRun/core-maintainers in case anyone else has a different opinion

jglovier:
>cc @HospitalRun/core-maintainers in case anyone else has a different opinion

No, I agree @jkleinsc. Giving admins a way to manually set this for patients seems like the best path forward at this point. 👍

pete-the-pete:
Setting that status _is_ more straightforward.  I filtered the data in the route's model hook.  I couldn't get the filterBy property to actually accomplish things (I'd like to talk that through in slack and some point).

Hopefully this is attempt is closer.

jkleinsc:
@pete-the-pete to use the filterBy parameter on the couch view, you can pass it along by overriding the _modelQueryParams function in the route.  For example:
```javascript
  _modelQueryParams() {
    let queryParams = this._super(...arguments);
    queryParams.filterBy = [{
      name: 'status',
      value: 'Missed'
    }];
  },
```

This will pass along the filter value to couch view.

pete-the-pete:
@jkleinsc thanks for the help.  I ignored the cors errors, and saw it working.  Hopefully this is it.

jkleinsc:
@pete-the-pete looks good... just one more thing.  The controller and template for missed should just extend/use the controller and template from app/appointments/index because the code exactly them same.  If you take a look at app/appointments/today/controller.js and app/appointments/today/template.hbs you will see how that route uses the index controller and template.  If you don't have time to do this let me know and I'll do it.

pete-the-pete:
I updated the controller and template.  Hopefully its good to go.

PR #404, opened 2016-04-05
Added ability to save roles

### LINKED ISSUE #338: Allow custom roles
Right now all of the roles in HosptialRun are hard coded to certain capabilities as defined in https://github.com/HospitalRun/hospitalrun-frontend/blob/master/app/mixins/user-session.js.

We should add the capability of defining custom roles and/or modifying the capabilities of predefined roles.

jkleinsc:
Fixes #338 .

This PR adds a new admin screen to customize the user roles in the system.  If a user role has not been customized, it will use the default values defined for each role.  If a role has been customized, any users logging in with the customized role will have access to the capabilities selected on the admin screen.
Here is a screenshot of the admin screen:
screen shot 2016-04-05 at 5 20 38 pm
-

@jglovier I think I showed you this, but wanted to get your blessing before we merge this PR.  I need to ship this feature this week for the Philippines.


jglovier:
@jkleinsc :cool:. It definitely needs some ❤️, but given that a) you need to ship it this week for implementation, and b) a lot of the :heart: it needs will be covered by more global form work, I'm 👍 with shipping as is for now.

PR #411, opened 2016-04-08
added location type to visit

karifa-kouyate:
Fixes # .

added more precision to visit location. with a drop down for the location type with  three options : 
.Health facility,
.Home visit,
.Other.
If you choose Health Facility a visit location field appears to enter the facility location,
if you choose Home visit that option is for home visit cases,
and if the location is  different from Heath facility and home visit you can choose third option Other and fill in the Other location field.
## 

### DIFF
--- a/app/models/visit.js
+++ b/app/models/visit.js
@@ -32,6 +32,8 @@ export default AbstractModel.extend({
   imaging: DS.hasMany('imaging', { async: true }),
   labs: DS.hasMany('lab', { async: true }),
   location: DS.attr('string'),
+  visitLocationType: DS.attr('string', {defaultValue:'Health Facility'}),
+  visitLocationOther: DS.attr('string'),
   medication: DS.hasMany('medication', { async: true }),
   // this field is being deprecated in favor of patient-note
   notes: DS.attr('string'),
--- a/app/visits/edit/controller.js
+++ b/app/visits/edit/controller.js
@@ -125,6 +125,24 @@ export default AbstractEditController.extend(ChargeActions, PatientSubmodule, Pa
     'Admitted',
     'Discharged'
   ].map(SelectValues.selectValuesMap),
+  
+  visitLocationTypes: [
+    'Health Facility',
+    'Home Visit',
+    'Other'
+  ].map(SelectValues.selectValuesMap),
+  
+  isLocationFacility: function() {
+    var visitLocationType = this.get('model.visitLocationType'),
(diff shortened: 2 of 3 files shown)


jglovier:
@karifa-kouyate thanks for the pull request! If these changes address or resolve an open issue, please reference it in the `Fixes # .` line. Otherwise you can delete that line.

jkleinsc:
@karifa-kouyate Can you explain further why you need this field versus the location field that is already there?  The current location field by default is a smart typeahead field (where it remembers previous entries), but it can also be configured as a drop down list.

karifa-kouyate:
@jkleinsc  The doctor can have visits outside the Health Facility , like Home-visit or other locations, so the visit type is added to be able to categorize them in case that we can know how many Home-visit,Health Facility or other locations visits we have.

tangollama:
Thanks for the clarification. That *feels* to me like the location field could accommodate that. One of the design principles were trying to embrace is simplicity. So unless there's a strong reason for a multi-level taxonomy, I'd resist adding this field. 

That said, the best idea wins and even though we're embracing simplicity, we also want to do what makes sense. With that backdrop, can you offer up a strong use case or argument for offering a multi-leveled taxonomy for location vs just storing all location values is a single, flat field.

jkleinsc:
@karifa-kouyate can you offer a strong use case or argument for offering a multi-leveled taxonomy for location vs just storing all location values is a single, flat field?  If not, we are going to close this PR.

douno:
Hello @jkleinsc , I work with @karifa-kouyate . The reason we multi-leveled taxonomy for location is because not all locations are the same type. It's important to be able to differentiate between a health center, or home visit. We felt that putting all of them in one field would make aggregation harder and could confuse the user of the app. It's also important to know that just an autocomplete field will confuse some of the users in smaller communities where those users are not well versed in basic IT, as has been the case for us.

jkleinsc:
@douno in regards to the autocomplete field, the existing location field can be configured to just be a drop down by going to Administration/Lookup Lists/Visit Locations and making sure the checkbox labeled "User Can Add New Values" is not checked.  This of course wouldn't give you the ability to pick from the drop down OR type in a value under other, but it does make things simpler.

tangollama:
@douno are you planning to continue work on this PR or should we close it? Let us know.

jkleinsc:
@tangollama I am going to close this issue.  Location as it stands right now is pretty robust and if folks need an alternate, they can do so now with custom forms.

PR #427, opened 2016-04-19
Nitrous.io Quickstart button

yeouchien:
- Add Nitrous Quickstart button

Added Nitrous Quickstart files so anyone can quickly spin up a free, cloud dev environment to start using this repo. Clicking the Quickstart button will automatically clone this repo and will install and configure any required packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,15 @@ To install the frontend please do the following:
         2. If you already have a CouchDB admin user, please run `./script/initcouch.sh USER PASS` in the folder you cloned the HospitalRun repo.  `USER` and `PASS` are the CouchDB admin user credentials.
 8. Copy the `server/config-example.js` to `server/config.js` in the folder you cloned the HospitalRun repo.
 
+### Nitrous Quickstart
+Create a free development environment for this HospitalRun project in the cloud on [Nitrous.io](https://www.nitrous.io) by clicking the button below.
+
+<a href="https://www.nitrous.io/quickstart">
+  <img src="https://nitrous-image-icons.s3.amazonaws.com/quickstart.png" alt="Nitrous Quickstart" width=142 height=34>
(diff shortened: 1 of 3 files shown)


jkleinsc:
@yeouchien thanks for the PR!  

I ran into a couple of issues testing this:
1) I didn't have a nitrous.io account and after I created one, it didn't create a project, but it did create a system.  This could be because I clicked on the checkbox to create an organization when I created my user but then I bailed out of that process
2) I tried again once I had a nitrous.io account and clicking the button successfully created the project, but the NPM install failed as well as the rest of the install.  I was able to successfully get the app running by running the install scripts manually, but it would be great if it works right out of the box.
3) I see that this is setup as a Rails project, which is misleading since it is a Node/Javascript project.  There is a gem that needs to be installed so maybe that is why this is setup as a rails project?
4) I was disappointed to see the free version only comes with 50 hours of coding time.  Cloud9, whom I assume is a competitor of yours offers free workspaces without limit.  I understand this is a business decision, but it would be awesome at least for open source projects that you lift this limit and/or make it substantially larger.

yeouchien:
Hi @jkleinsc,

> I clicked on the checkbox to create an organization when I created my user but then I bailed out of that process

Yea currently `quickstart` context will be dropped when a new user `create an organization`.

> NPM install failed

NPM failed randomly when I tried to manually setup this project in Nitrous. I've tested a few times with no NPM install fail. So, I guess this is just some random error from NPM.

> this is setup as a Rails project

I'm using `Ruby` template in this project. It has `NodeJS` installed in the project container too. The reason I choose `Ruby` is because this project need Ruby for scss linter -- `Make sure you have Ruby installed (for the scss linter).`

Thanks for your time for this testing out this PR! 

Issue #397, opened 2016-03-28
'No imaging found' helper link not presented after removing all imaging


brianlindsy:
**Expected behavior:**
When you remove all imagning, you should get the helper starting link back

**Actual behavior:**
The helper link is not present

**Steps to reproduce:**
Be in the state with only one image
Delete that image
The 'No imaging found. Create a new imaging record' starting helper is not presented.
**Screenshots (if applicable):**

**OS and Browser:**

rohansp:
Is there an option to delete an imaging request?
I can see only the edit button
![imaging request](https://cloud.githubusercontent.com/assets/11171608/14130650/336b0bc4-f5fa-11e5-896c-3c9985d24e3c.png)

tangollama:
@rohansp you can fulfill the request to empty the queue.

Issue #396, opened 2016-03-28
'No appointments found' helper link not presented after removing all appointments


brianlindsy:
**Expected behavior:**
When you remove all appointments, you should get the helper starting link back

**Actual behavior:**
The helper link is not present.

**Steps to reproduce:**
Be in the state with only one appointment
Delete that appointment
The 'No appointments found. Create a new appointment' starting helper is not presented.
**Screenshots (if applicable):**

**OS and Browser:**

PR #431, opened 2016-04-22
Fix #395 #396 #397

### LINKED ISSUE #395: 'No items found' helper link not presented after removing all items
**Expected behavior:**
When you remove all items, you should get the helper starting link back

**Actual behavior:**
Helper link is not present

**Steps to reproduce:**
Be in the state with only one item
Delete that item
The 'No items found. Create a new item record' starting helper is not presented.

**OS and Browser:**

esbanarango:
Fixes #395, #396 and #397.

- Remove the assignation of property `hasRecords` on the `setupController` of `abstract-index-route`.
- Add a computed property `hasRecords` (based on `model.length`) to `abstract-paged-controller`. This way when there are no more items this property will get updated.


jkleinsc:
@esbanarango looks good.  Thanks for the PR!

PR #437, opened 2016-04-29
Add `Requested By` to Imaging

### LINKED ISSUE #362: Requesting clinician needs to be added to the Imaging Request list
The request list does not contain the requesting user (clinician). We need to see this for the purposes of ensuring the person processing the requests that they are appropriately authorized. Even though we permission specific people to issue those requests, we still want the user experience to verify that.

cacqw7:
Fixes #362.

- Adds `Requested By` to Imaging **Requests** and **Completed** lists
- Refactors Imaging to use [components in blocks](https://github.com/DockYard/styleguides/blob/master/engineering/ember.md#use-components-in-each-blocks)
- Adds `Add Note` link to the **Requests** list items that don't have notes
- Adds `No Notes` and `No Results` to **Completed** list items that don't have either Notes or Results

---

`Add Notes` link:
link

`No Notes` and `No Results`
light

### DIFF
--- a/app/styles/components/_imaging.scss
+++ b/app/styles/components/_imaging.scss
@@ -1,0 +1,1 @@
+.notes { max-width: 300px; }
--- a/app/imaging/completed/completed-list-item/component.js
+++ b/app/imaging/completed/completed-list-item/component.js
@@ -1,0 +1,5 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  tagName: 'tr'
+});
--- a/app/imaging/completed/completed-list-item/template.hbs
+++ b/app/imaging/completed/completed-list-item/template.hbs
@@ -1,0 +1,7 @@
+<td>{{date-format imaging.requestedDate}}</td>
+<td>{{date-format imaging.imagingDate}}</td>
+<td>{{imaging.requestedBy}}</td>
+<td>{{imaging.patient.displayName}}</td>
+<td>{{imaging.imagingType.name}}</td>
+<td>{{imaging.results}}</td>
+<td>{{imaging.notes}}</td>
--- a/app/imaging/completed/completed-list-item/template.hbs
+++ b/app/imaging/completed/completed-list-item/template.hbs
@@ -1,7 +1,7 @@
-<td>{{date-format imaging.requestedDate}}</td>
-<td>{{date-format imaging.imagingDate}}</td>
-<td>{{imaging.requestedBy}}</td>
-<td>{{imaging.patient.displayName}}</td>
-<td>{{imaging.imagingType.name}}</td>
-<td>{{imaging.results}}</td>
-<td>{{imaging.notes}}</td>
+<td class="requested-date">{{date-format imaging.requestedDate}}</td>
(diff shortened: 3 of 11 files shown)


jkleinsc:
@cacqw7 thanks for the PR!
A couple of things:
1) Can you change your code to use localization for the text displayed?  More details about localization using ember-i18n can be found here: https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Translating-Text
2) Can you provide screenshots of the change you made.

cacqw7:
@jkleinsc done and done!

jkleinsc:
@cacqw7 thanks for the update!

@jglovier Any objections/comments on the UI changes for this PR?

jglovier:
Right now there is a lot of wrapping of very short fields. It'd be better if we could shorten up the Notes field (maybe `max-width: 300px`?) and give space for the other fields to not wrap so easily.

jglovier:
Sorry, entered that last comment before finished...

- I'm not a fan of the "No results" and "No notes" placeholder text because it's redundant and unnecessary, and adds visual noise. If there are no notes to display, I think the absence of them is clue enough to that.
- I like surfacing the `Add notes` link, but I'm curious to know more of your intentions there. Does it only appear on hover, or visible by default on every item without notes? I'd be opposed to adding it on every single item w/o notes for the same visual noise concerns as above, even if it is more useful

Unfortunately some of this just comes back to incomplete work on how actions are surfaced on items in the first place. I don't want to block too much here on account of that, but those are important concerns to resolve before we go too far with implementing more pieces that will be addressed by that.

Thanks for jumping in here @cacqw7! :sparkles:

cacqw7:
@jglovier 

I'll go ahead and remove `No Results`, `No Notes` _and_ the `Add Notes`. Those were all outside the scope of #362 and I agree that there are more important things to resolve first. 

I'll work on the margins this weekend and push up changes/screenshots 👍 😄

jglovier:
@cacqw7 awesome. :metal: As we get closer to resolving the surfacing actions in index views piece, I'll rope you back in and we can revisit these pieces under the new paradigm.

cacqw7:
@jglovier I added the `min-width` you suggested. 

request
min-width

app/styles/components/_imaging.scss
```diff
@@ -0,0 +1,2 @@
+.imaging-row .notes { max-width: 300px; }
+.completed-imaging-row .notes { max-width: 300px; }
```
jkleinsc: @cacqw7 not sure why we need two different classes here.  @jglovier any opinion on using two different classes for completed vs in-progress?

jglovier: Yeah, if the value is the same (which it appears to be) and there is no other conflict, just use the existing classname.

jglovier:
>I added the min-width you suggested.

@cacqw7 that looks much better. I'd love to add a min-width for the patient name, too. But that's probably more of a global concern that we can address in another PR.

cacqw7:
@jkleinsc I removed those classes 👍

jkleinsc:
Looks good @cacqw7.  I am going to merge it in.

PR #474, opened 2016-05-20
Corrected one of the step numbers in the "Make" section

vanakema:
I assume a step was added between when the “Make” section was last
updated and now, so I corrected the step numbers that must be completed.


jkleinsc:
Thanks for the PR @vanakema !

PR #486, opened 2016-06-02
Test for missing translations

billybonks:
Stubs the translation function that do the translation and
throw an error if it is missing

uses deprecated api, happy to change but atleast it is only one
instance

All Feature tests must run startApp in the before each hook


jkleinsc:
Looks good @billybonks.  I'm ok using the deprecated api for now.

billybonks:
Coool thanks :)

PR #509, opened 2016-06-16
Group stylelint rules by type

billybonks:
Groups stylelint rules by type. as their documents recommend

should we order alphabetically ?


jkleinsc:
Looks good @billybonks.  I think the alphabetical /by type makes sense.

PR #523, opened 2016-06-20
Update ember-cli-stylelint

billybonks:
@jglovier @jkleinsc 

fixes the issue where the stylelint plugin was breaking live reload.


jkleinsc:
Looks good @billybonks.

PR #538, opened 2016-06-29
Fixes Issue 467

### LINKED ISSUE #467: Billed Invoices - Buttons disappear from the invoice
**Expected behavior:**
Buttons to add payment, edit, delete  are always there.

**Actual behavior:**
If you click on add payment, the buttons disappear.   

**Steps to reproduce:**
on beta site
Billing > Invoices > click on the add payment button for an invoice.   The buttons disappear when the dialog comes up.   If you dismiss the dialog with cancel, the buttons are gone.

screen shot 2016-05-13 at 4 21 58 pm

**OS and Browser:**
OSX   Chrome

kaungst:
Fixes #467 

**Problem Summary**
- A `payment` model is created when hitting the `Add Payment` button on invoices. 
- This model is invalid, but it's creation sets the `paidTotal` to 0. 
- This set's `status` to 'Paid' for invoices requiring 0 payment.

**Fix Summary**
- request adds additional logic to where the `status` field is set
  - keeps track of last unpaid status if it needs to return to it
    (e.g. return to either Bill or Draft)
- fixes missing checks for invalid payments when calculating remaining
  balance
- implements cleanup function to destroy payment model placeholder
  upon modal exit with invalid model state.

### DIFF
--- a/app/invoices/payment/controller.js
+++ b/app/invoices/payment/controller.js
@@ -3,10 +3,22 @@ import Ember from 'ember';
 import PatientSubmodule from 'hospitalrun/mixins/patient-submodule';
 
 export default AbstractEditController.extend(PatientSubmodule, {
+  closeModalAction: 'cleanup',
   cancelAction: 'closeModal',
   findPatientVisits: false,
   invoiceController: Ember.inject.controller('invoices'),
   newPayment: false,
+
+  actions: {
+    cleanup: function() {
+      var model = this.model;
+      this.model.validate().then(function() {
+        model.save();
+      }).catch(function() {
+        model.destroyRecord();
+      });
+    }
+  },
 
   expenseAccountList: Ember.computed.alias('invoiceController.expenseAccountList'),
   patientList: Ember.computed.alias('invoiceController.patientList'),
--- a/app/models/invoice.js
+++ b/app/models/invoice.js
@@ -14,6 +14,7 @@ export default AbstractModel.extend(DateFormat, NumberFormat, {
   visit: DS.belongsTo('visit', {
     async: false
   }),
+  lastUnpaidStatus: DS.attr('string'),
   status: DS.attr('string'),
   remarks: DS.attr('string'),
   billDate: DS.attr('date'),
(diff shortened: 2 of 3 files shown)


app/invoices/payment/controller.js
```diff
@@ -3,11 +3,23 @@ import Ember from 'ember';
...
   cancelAction: 'closeModal',
   findPatientVisits: false,
   invoiceController: Ember.inject.controller('invoices'),
   newPayment: false,
 
+  actions: {
+    cleanup: function() {
+      var model = this.model;
+      this.model.validate().then(function() {
+        model.save();
```
jkleinsc: This cleanup code should not save the model because the model should not be saved on the cancel action, it should only be saved when the user clicks on the update/add action which is already handled elsewhere in the controller.

app/models/invoice.js
```diff
@@ -14,6 +14,7 @@ export default AbstractModel.extend(DateFormat, NumberFormat, {
   visit: DS.belongsTo('visit', {
     async: false
   }),
+  lastUnpaidStatus: DS.attr('string'),
```
jkleinsc: I don't think this status is necessary.  See code comments below.

app/models/invoice.js
```diff
@@ -108,14 +109,23 @@ export default AbstractModel.extend(DateFormat, NumberFormat, {
   patientResponsibility: Ember.computed.sum('patientResponsibilityTotals'),
 
   paymentAmountChanged: function() {
-    var payments = this.get('payments'),
-      paidTotal = payments.reduce(function(previousValue, payment) {
+    var payments = this.get('payments');
```
jkleinsc: I think this code block could be simplified to just keep track of if the payments are valid and then set the status to paid if remainingBalance <=0 and hasValidPayments === true.  There really isn't a need to do anything if those conditions are untrue.  There shouldn't be a need to reset the status back to the status it is already set to.

jkleinsc:
@kaungst Thanks for the PR.  I made some comments in your code changes.

tangollama:
@kaungst, hi. Are you planning to address this PR further or should we close it?

tangollama:
It looks like this issue has been addressed in pr #654. @jkleinsc can you confirm and close if so?

jkleinsc:
@tangollama yup.  Closing this one.

PR #539, opened 2016-06-30
added feature #460 option to create a new visit on the creation of a …

### LINKED ISSUE #46: Make Visit Type an editable lookup list
Philippines needs a Physical Therapy visit type.  We should just make Visit Type an editable lookup list

### LINKED ISSUE #460: Add an option to create a new Visit on the creation of a new Patient
Include at the bottom of the New Patient screen a checkbox (by default unchecked) to automatically create an Outpatient Visit for this Patient. If unchecked, maintain current behavior. If checked, create the Visit after the creation of the Patient record and navigate to the New Visit screen.

This will aid in the admission process.

omoabobade:
Fixes #460

- added a checkbox with property "createOutpatientVisit" to the add patient form
- the checkbox is uncheck by default
- added property "createOutpatientVisit" in the patient model
- added a check to see if checkbox is checked then add new visit and redirect to new visit page

_Note: pull requests without proper descriptions may simply be closed without further discussion. We appreciate your contributions, but need to know what you are offering in clearly described format. Thanks! (you can delete this text)_

…new Patient

### DIFF
--- a/app/models/patient.js
+++ b/app/models/patient.js
@@ -15,6 +15,7 @@ export default AbstractModel.extend(DOBDays, PatientName, {
   bloodType: DS.attr('string'),
   clinic: DS.attr('string'),
   country: DS.attr('string'),
+  createOutpatientVisit: DS.attr('boolean'),
   dateOfBirth: DS.attr('date'),
   economicClassification: DS.attr('string'),
   email: DS.attr('string'),
--- a/app/models/patient.js
+++ b/app/models/patient.js
@@ -15,7 +15,6 @@ export default AbstractModel.extend(DOBDays, PatientName, {
   bloodType: DS.attr('string'),
   clinic: DS.attr('string'),
   country: DS.attr('string'),
-  createOutpatientVisit: DS.attr('boolean'),
   dateOfBirth: DS.attr('date'),
   economicClassification: DS.attr('string'),
   email: DS.attr('string'),
--- a/app/patients/edit/controller.js
+++ b/app/patients/edit/controller.js
@@ -517,6 +517,12 @@ export default AbstractEditController.extend(BloodTypes, ReturnTo, UserSession, 
   },
 
   afterUpdate: function(record) {
+    var patient = this.get('model');
+    if (patient.getProperties('createOutpatientVisit').createOutpatientVisit) {
+      var visits = this.get('model.visits');
+      this.send('createNewVisit', patient, visits);
+      this.transitionToRoute('visits.edit');
+    }
     this.send('openModal', 'dialog', Ember.Object.create({
       title: this.get('i18n').t('patients.titles.saved_patient'),
       message: this.get('i18n').t('patients.messages.saved_patient', record),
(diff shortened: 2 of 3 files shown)


app/models/patient.js
```diff
@@ -15,6 +15,7 @@ export default AbstractModel.extend(DOBDays, PatientName, {
   bloodType: DS.attr('string'),
   clinic: DS.attr('string'),
   country: DS.attr('string'),
+  createOutpatientVisit: DS.attr('boolean'),
```
jkleinsc: This attribute should not be here because we do not want to (or need to) save this value to the database.

app/patients/edit/controller.js
```diff
@@ -517,6 +517,12 @@ export default AbstractEditController.extend(BloodTypes, ReturnTo, UserSession,
   },
 
   afterUpdate: function(record) {
+    var patient = this.get('model');
+    if (patient.getProperties('createOutpatientVisit').createOutpatientVisit) {
```
jkleinsc: instead of
 ```javascript
if (patient.getProperties('createOutpatientVisit').createOutpatientVisit) {
```
You should use patient.get. For example:
 ```javascript
if (patient.get('createOutpatientVisit')) {
```

jkleinsc:
@omoabobade Thank you for your PR!  I have made some comments above about the change you made.  Please take a look at them.  Also, I see that the tests are not passing and it appears the reason is because when you create a new patient it doesn't have any visits set which causes the errors.  If you change your code to only work on new records, we can fix this error.  I'll add more comments to your code to show you what needs to be done.

app/patients/edit/template.hbs
```diff
@@ -147,6 +147,12 @@
                     {{ext-radio content=patientTypes value=model.patientType showInline=true }}
                   </div>
                 </div>
+                <div class="form-input-group">
```
jkleinsc: This block should be wrapped with an if that only displays this checkbox if the patient is new.  Eg:
```handlebars
{{#if model.isNew}}
  <div class="form-input-group">
    <label>Create An Outpatient Visit</label>
    <div>
      {{em-checkbox property="createOutpatientVisit" }}
      </div>
  </div>
{{/if}}
```

app/patients/edit/controller.js
```diff
@@ -517,6 +517,12 @@ export default AbstractEditController.extend(BloodTypes, ReturnTo, UserSession,
...
+      var visits = this.get('model.visits');
```
jkleinsc: visits is always going to be null in this situation, so instead of
```javascript
      var visits = this.get('model.visits');
```
use this:
```javascript
      var visits = [];
```

This will fix the error thrown in the tests.

jkleinsc:
Also @omoabobade once you have made the suggested changes, you should run ```ember test``` in your development environment to verify all the tests are passing.  Alternatively you can run ```ember serve``` and then go to [http://localhost:4200/tests](http://localhost:4200/tests) to run the tests in your browser.

tangollama:
@jkleinsc does this just need to re-run the tests or this there a gap in the functionality that @omoabobade submitted?

jkleinsc:
@tangollama there are errors in the submitted PR that need to be corrected.  I highlighted above what those errors are and how to correct them,

jkleinsc:
Closing this PR in favor of merging in outpatient branch.

PR #544, opened 2016-07-01
Add translations for ember-validations

billybonks:
Required strings for ember-validations to be translated as well 👍 


jkleinsc:
Looks good as well.  Going to merge.

PR #545, opened 2016-07-01
remove <strong> from delete message

billybonks:
I previously let `<strong>` tag in delete message through to master. after reading all delete strings this does not seem to be the standard so i removed the tags. 


jkleinsc:
Looks good.  I'll merge this.

PR #550, opened 2016-07-03
Use super

billybonks:
We should use the super method when available instead of rewriting code. also changed update button for admin section to follow the same pattern as the rest of the application


jkleinsc:
Looks good.  I'll merge.

PR #558, opened 2016-07-08
Updating sortable columns to have sorting icon and updated cursor

### LINKED ISSUE #328: Need Indicator for Sortable Columns
On the /inventory/listing page, there are a bunch of sortable columns to view the inventory and sort accordingly. However, prior to clicking on the column header, there is no indication that you can sort the column. Even on mouse over, the cursor is the text-selector and not the link-clicker. Before sorting After sorting

I would recommend adding an icon like [fa-sort](http://fortawesome.github.io/Font-Awesome/icon/sort/) as part of each sortable-column. Other sorting icons can be found [here](http://fortawesome.github.io/Font-Awesome/icons/) and searching with "sort" as the search phrase.

pete-the-pete:
Fixes #328 

- add new icon for default state of sortable-column
- add 'sortable-column' classname to sortable-column component
- add a new css rule to use the pointer cursor for .sortable-column class

sorting icons

### DIFF
--- a/app/components/sortable-column.js
+++ b/app/components/sortable-column.js
@@ -1,5 +1,6 @@
 import Ember from 'ember';
 export default Ember.Component.extend({
+  classNames: ['sortable-column'],
   tagName: 'th',
   action: 'sortByKey',
   sortDesc: false,
--- a/app/styles/_temp_misc.scss
+++ b/app/styles/_temp_misc.scss
@@ -1,8 +1,17 @@
 $table-header-color: rgba($navy_drk2,.7);
+$inactive-sort-indicator: rgba($navy_drk2,.3);
 
 .table-header {
   background-color: $navy_light;
   color: $table-header-color;
+
+  .sortable-column {
+    cursor: pointer;
+
+    .inactive {
+      color: $inactive-sort-indicator;
+    }
+  }
 }
 
 .twitter-typeahead {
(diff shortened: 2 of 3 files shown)


jglovier:

jglovier:
@pete-the-pete thanks for getting started on this! Only design feedback I have at the moment is we should probably reduce the size of the icons a bit, and instead of the double arrows, lets use a single triangle everywhere of [this style](https://octicons.github.com/icon/triangle-down/). To differentiate between active filtering and non-active, lets take a queue from existing patterns and make the triangle color lighter on the non-active filters, and darker on active filters.

jglovier:
If you'd like visual help beyond my written description above, I can post a design comp when I have some time in the next few days.

pete-the-pete:
I can take a look tomorrow and post an update.

pete-the-pete:
I updated to use the arrows and lighter color, but I am seeing this issue, https://github.com/stylelint/stylelint/issues/1513

jkleinsc:
@pete-the-pete looks good to me.  @jglovier what do you think:

screen shot 2016-07-11 at 2 51 16 pm

jglovier:
The new icons looks good, except I think we should only display the active arrow, and hide the others until you hover over it's associated column heading.

jkleinsc:
@jglovier if you look the issue this PR is tied to, #328, it seems the issue is discoverability.  If we don't show any icons, how will users know they can search?

jkleinsc:
@jglovier ping... what do you think?  I would like to move this PR along.

jglovier:
@jkleinsc, sorry, to be more clear I think by default we should always show an icon for the currently active sorted column. So there is always one sorting icon visible.

Also, I'm not too concerned about discoverability because this is a feature most users will probably assume is there already. Sorting tabular data by column is more of an expected behavior than a novel feature we need to explicitly reveal.

jkleinsc:
I am merging this PR as is.  We can cleanup when the icons show as another issue.

PR #559, opened 2016-07-08
Translate printed invoice

billybonks:
Translates invoices. currently ignoring the template because we aren't translating specific to hospital strings.


jkleinsc:
@billybonks we need to take this string out of the translations:
hospital_footer: 'Certified Correct by: Leron L. Lehman- Executive Director',

billybonks:
@jkleinsc updated 👍

jkleinsc:
Thanks @billybonks.  I'll merge.

PR #561, opened 2016-07-08
Translate/missed strings

billybonks:
A couple of strings where missed, they are translated in this pr


jkleinsc:
@billybonks this PR uses some of the old underscore convention -- Do you want me to merge this PR as is or do you want to remove/change the underscores first?  I'm fine either way.

billybonks:
you are correct may as well start using the proper conventions. right away. made the updates

PR #570, opened 2016-07-13
README Fix text flow issue around screenshot

devinrhode2:
"You should see the following screen:" gets cut off at "You should see" with the screenshot in between, and "the following screen:" shown to the bottom left of this screenshot.

Adding an extra newline fixes the issue.


billybonks:
LGTM 👍

PR #595, opened 2016-07-17
Added scrollbar and more options in typeahead

### LINKED ISSUE #519: Medication/inventory typeahead issues
**Expected behavior:**
If multiple values matched the typed in value, they should all display, perhaps via a scrollbar
**Actual behavior:**
Unless you start typing more specific text, the typeahead just picks the first 5 matches without scrolling.  
**Steps to reproduce:**
Go to medication/new request and type in a single letter into the medication type ahead
screen shot 2016-06-18 at 5 42 04 pm
The first typeahead didn't show "Hyoscine 10mg tablet (Buscopan) - m00177 (104 available)"
screen shot 2016-06-18 at 5 41 42 pm

**OS and Browser:**
All

cmwebby:
Fixes #519 Medication/inventory typeahead issues 

- Typeahead now displays a maximum  of 10 options by adding limit in typeahead.js. Adding more than 10 appears to make the acceptance tests timeout
- Typeahead now has a maximum height of 110px and displays a vertical scrollbar 

# Example
typeahead 2

### DIFF
--- a/app/components/type-ahead.js
+++ b/app/components/type-ahead.js
@@ -45,6 +45,7 @@ export default InputComponent.extend({
   hint: true,
   highlight: true,
   lastHint: null,
+  limit: 10,
   minlength: 1,
   selectedItem: false,
   inputElement: null,
@@ -73,6 +74,7 @@ export default InputComponent.extend({
       minLength: this.get('minlength')
     }, {
       displayKey: this.get('displayKey'),
+      limit: this.get('limit'),
       source: this._getSource(),
       templates: this.get('templates')
     });
--- a/app/components/type-ahead.js
+++ b/app/components/type-ahead.js
@@ -45,7 +45,7 @@ export default InputComponent.extend({
   hint: true,
   highlight: true,
   lastHint: null,
-  limit: 10,
+  limit: 500,
   minlength: 1,
   selectedItem: false,
   inputElement: null,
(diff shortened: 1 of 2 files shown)


jkleinsc:
@cmwebby - thanks for the PR!  Quick question - instead of limiting to 10 can we have it be unlimited with the scrollbar?  Maybe 10 is good enough, but just curious what you found.

cmwebby:
@jkleinsc Firstly can I apologise for making then closing a like a million pull requests! I don't normally work this way!
I think it certainly could be made to be unlimited but I think there may be a couple of issues with this:

1. The main issue is whenever I tried adding a number more than 10 it made all the acceptance tests timeout. I couldn't spot the reason why it did this so I just stuck with 10. 
e.g
> not ok 74 PhantomJS 2.1 - Acceptance | medication: returning medication
    ---
        actual: >
            null
        message: >
            Test timed out
        Log: |

2. If there are too many results say more than 50 it would probably just be quicker and easier for the user to just to type a few more characters in than searching all the possible results using the scroll bar. 

If you have any ideas of why it would make the tests fail I am happy to go back and increae the limit.

cmwebby:
@jkleinsc Ignore my previous comment for some reason the tests are passing now with more than 10. I have set a limit of 500, but it can be set to more if you want.

jkleinsc:
@cmwebby Looks good to me.  I just tested it with a large dataset and the typeahead had no problems loading 500 suggestions, so I think we will stick with that.  Thanks for the PR!

Issue #592, opened 2016-07-15
Translation Keys Procedures namespace


billybonks:
Please read tracking issue blurb #573

this issue is for the `procedures` namespace

Issue #591, opened 2016-07-15
Translation Keys Print namespace


billybonks:
Please read tracking issue blurb #573

this issue is for the `print` namespace

Issue #590, opened 2016-07-15
Translation Keys Billing namespace


billybonks:
Please read tracking issue blurb #573

this issue is for the`billing` namespace

PR #599, opened 2016-07-19
Fixes Issues 590, 591, 592

dapierce:
Fixes
* #590 
* #591 
* #592 

* The start of changing Translation keys from _ standard to camel case to work with JavaScript convention.

Additionally, this PR also fixes a problem I noticed while testing the invoice--previously the print style did not fix the left margin for the printed invoice, so I have added a fix for the invoice in _print.scss


billybonks:
Hey David thanks :) looks awesome can you add screen shots for the change you made to the printed invoice 👍 

@jkleinsc besides the css i think as long as these commits pass tests it's good to go  since we have the missing translation explosion.

dapierce:
I'll make a screenshot later today for the print invoice.

billybonks:
if you also have the bandwidth, changing the keys in the other translation files would help.

dapierce:
Here's the invoice printing with the improved stylesheet:

invoice print screenshot

I realized that the invoice uses a different class (panel-primary) for its div than the other pages in HospitalRun, so I added it to the print style positioning.

jkleinsc:
Looks good to me.  Thanks @dapierce for the PR!

Issue #596, opened 2016-07-18
Linters


charlespeters:
the simple long loved jscs is being end of life'd and consolidating with eslint. this project is using dockyard's ember-suave. they've started working on an [eslint port](https://github.com/DockYard/eslint-plugin-ember-suave) of it; it's still experimental but worth trying. 

**Expected behavior:**  keeping the code style focused and unified

**Actual behavior:** deprecation warnings

**Steps to reproduce:** being dumb enough (like me) to update your global installation of jscs.

**OS and Browser:** os x 10.11.5 and chrome (atom/vim)

jkleinsc:
@charlespeters are you still interesting in working on this?

charlespeters:
yep will fix tonight!

PR #601, opened 2016-07-20
adding eslint to project

charlespeters:
Addresses #596.

- adds eslint to project
- jscs rules don't match one for one to eslint so that isn't an exact match for current linting set so i'm hoping for feedback on this PR

### Other considerations

- removing jscs
- [disabling jshint](https://github.com/ember-cli/ember-cli-eslint#disabling-jshint) in favor of eslint


jkleinsc:
@charlespeters thanks for the PR!  Looking at https://github.com/DockYard/eslint-plugin-ember-suave it appears that it is still in progress and I would like to wait until that is stable to move over.  I'm ok leaving this PR open until that happens.

jkleinsc:
@charlespeters it looks like https://github.com/DockYard/eslint-plugin-ember-suave is ready to use:
https://dockyard.com/blog/2016/09/30/retiring-ember-suave

Would you be willing to update this PR so that we can move from ember-suave?  Thanks!

charlespeters:
@jkleinsc now the fun part of this is that now we're using eslint, some of the tests are failing.

PR #615, opened 2016-08-03
Minor README updates

unobe:
- Update README to fix missing parenthesis
- Update README to add note for Mac OS users to run CouchDB


jkleinsc:
@unobe Thanks for the PR.  I am concerned about adding the statement:

> (*Note: If on Mac, you need to make sure CouchDB can be run: go to System Preferences > Security & Privacy > Allow apps downloaded from: Anywhere*)

Because this is a system wide setting, not one particular to CouchDB, it essentially is suggesting to a user that they should be lax about the security of their OS.  I would prefer to recommend users just allow CouchDB specifically using instructions like the ones here under the section 
**How to open an app from a unidentified developer and exempt it from Gatekeeper**:
https://support.apple.com/en-us/HT202491

unobe:
@jkleinsc I've modified the patch to just refer to Apple's documentation: let the user understand what they want to do security-wise.

jkleinsc:
@unobe Looks good to me.  I'll merge it in.  Thanks again!

PR #645, opened 2016-09-11
Issue 627: Fix visit types list not refreshing.

### LINKED ISSUE #627: Visit Types cannot be changed; although they are in the Administration as changeable
**Expected behavior:**

When a new visit type is added in Administration > Lookup Lists, that new type is now available in the visit type drop-down on a visit page

**Actual behavior:**

Drop down list is not updated.

**Steps to reproduce:**

1. In Administration > Lookup Lists  select the drop down for visit type from the Lookup Type dropdown.  
2. click on Add Value and add a new visit type  ("testtype")
3. click on update, verify new type is in the list.
4. navigate to any patient record. 
5. Add a new visit, click on the visit type to reveal the drop down
6. new type is not shown

**OS and Browser:**

OSX , Chrome

akong:
Fixes #627 

- Fixes the typo that was causing the visitTypesList to not be computed properly.
- Add volatile to the dependent property of visitTypesList (see app/mixins/visit-types.js).  It looks like being dependent on a computer property doesn't quite work as any subsequent updates to the underlying data is not firing the observers (i.e. the mixin functions).  

### DIFF
--- a/app/appointments/edit/controller.js
+++ b/app/appointments/edit/controller.js
@@ -57,7 +57,7 @@ export default AbstractEditController.extend(AppointmentStatuses, PatientSubmodu
         isAdmissionAppointment = this.get('isAdmissionAppointment');
     return (!allDay && isAdmissionAppointment);
   }.property('model.allDay', 'isAdmissionAppointment'),
-  visitTypesList: Ember.computed.alias('appointmentsController.visitTypeList'),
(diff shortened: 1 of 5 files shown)


jkleinsc:
@akong thanks for the PR!  Looks good to me.  I'll merge it in.

PR #663, opened 2016-09-23
Hid checkbox labels

### LINKED ISSUE #322: Inconsistent Checkbox Spacings
There are a couple of forms in inventory that use checkboxes, but they look out of place on the page with their labels. For example, on the "+ new request" [page](https://github.com/HospitalRun/hospitalrun-frontend/blob/master/app/inventory/request/template.hbs), the "fulfill" checkbox seems to have a lot more whitespace then the rest of the page. When it's selected, a new "consumed" checkbox appears that also looks out of place in terms of element padding. The labels for the checkboxes seem to be too far away from the checkboxes as well.

This seems like it could be solved by either removing the labels for checkboxes, by adjusting the checkbox row height to be only half the other rows' heights, or by removing the bottom margins for the checkbox labels.

Image of New Request Page

jscottchapman:
Fixes #322.

-Hid redundant labels using “sr-only” class.
-The checkbox ended up where the label used to be so I added a class of
 “checkbox-label-space” that can be used where that comes into play.
gif:
http://recordit.co/qtt0q8Xwt7

### DIFF
--- a/app/inventory/batch/template.hbs
+++ b/app/inventory/batch/template.hbs
@@ -5,12 +5,14 @@
       {{select-or-typeahead property="vendor" label=(t 'inventory.labels.vendor') list=vendorList selection=model.vendor className="col-sm-4 required test-vendor"}}
       {{em-input property="invoiceNo" label=(t 'inventory.labels.invoiceNumber') class="col-sm-4 test-invoice-number"}}
     </div>
-    <div class="row">
+    <div class="form-group row">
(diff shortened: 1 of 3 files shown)


tangollama:
@jglovier look good to you?

jscottchapman:
@jglovier Merge?

jkleinsc:
@jscottchapman thanks for the PR.  I am going to merge it in.

Issue #572, opened 2016-07-15
Patient ID can be made the same


cmwebby:
**Expected behavior:**
The Patient ID should be completely unique to the patient. Although the database has its own unique id for each patient and the patient ID is just a user-friendly one, it should still be completely unique to avoid mixing up patients. 

**Actual behavior:**
When making a new patient the Patient ID can be edited to a patient ID  that already exists. 

**Steps to reproduce:**
- Make a new patient and take note of the ID
- Make another patient and edit the ID to be the same

**Screenshots (if applicable):**
![patient id](https://cloud.githubusercontent.com/assets/15896226/16869670/e755f944-4a75-11e6-9ad5-69004fdf3989.png)

**OS and Browser:**
OS - Windows 10
Browser - Version 51.0.2704.103 m

JSchneidler:
I'll see if I can tackle this. Would it be acceptable to just make an API request when the Id field is changed to check if it has already been used?

jkleinsc:
Thanks @JSchneidler !  If you look at app/mixins/patient-id, there is code there that generates this id on a new record via the `generateFriendlyId` function.  That function queries a couch view called `patient_by_display_id`.  You can query this same view to find out if an id is being used.

taras:
@JSchneidler are you able to look at this issue? If you are, I'd like to assign it to you. Otherwise, we'll assign it to someone else. 

Please, let us know if you're able to work on it.

gnowoel:
@taras Since I'm working on a related issue (#485) , I'd be happy to take this if no one else does.

mdarmani:
Is this assigned? I don't see anyone assigned yet.

jkleinsc:
@mdarmani @gnowoel is working on this.

yasirunilan:
is this one assigned? I would like to work on this.

jkleinsc:
@yasirunilan @gnowoel is working on this one.

mdarmani:
@jkleinsc Shouldn't they be assigned to this? I am looking for a Hackathon story to pick up. Or van we work on this as a trio?

jkleinsc:
@mdarmani I can't assign issues to outside contributors, so our workflow for issues is to remove the help wanted tag once someone starts working on it, so if you are looking for something to work on, take a look at the help wanted issues:
https://github.com/HospitalRun/hospitalrun-frontend/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22

[...]

PR #667, opened 2016-09-24
Fixes #485: Suggest next patient ID based on the largest number used

### LINKED ISSUE #485: Patient ID sequence should not go backwards (greater to lesser)
It was observed and confirmed that at the Tebow hospital, the patient ID sequence was set at a number less than the greatest patient ID in the system.   There were 2000+ patients in the database currently, but the sequence was set at something around 1400.   It was not causing an issue, since the system prevented creation of patients with duplicate IDs.   However, in consult with @jkleinsc it was agreed that this behavior should be corrected.

gnowoel:
The suggested patient ID keeps getting incremented even if we cancelled the new patient form. This problem should be related to issue #485 and more info can be found in https://github.com/HospitalRun/hospitalrun-frontend/issues/485#issuecomment-249384870.

I try to solve the problem by saving the _current_ sequence rather than the next available one. We start checking from the current sequence when preparing a new patient form. The value would be incremented and saved only if it has already been taken. So, no matter how many times we click the cancel button, the suggested patient ID would always be the same.

Please correct me!

### DIFF
--- a/app/mixins/patient-id.js
+++ b/app/mixins/patient-id.js
@@ -19,11 +19,11 @@ export default Ember.Mixin.create(PouchDbMixin, {
     const maxValue = this.get('maxValue');
 
     const findUnusedId = (sequence) => {
-      let next, id;
+      let current, id;
       return config.getPatientPrefix()
         .then(function(prefix) {
-          next = sequence.incrementProperty('value');
-          id = sequenceId(prefix, next);
+          current = sequence.get('value');
+          id = sequenceId(prefix, current);
           const query = {
             startkey: [ id, null ],
             endkey: [ id, maxValue ]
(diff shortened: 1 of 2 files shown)


gnowoel:
Added a [commit](https://github.com/HospitalRun/hospitalrun-frontend/pull/667/commits/df284994e3fd838b0c7a937e600e3f18f9e99d36) to solve the first half of the problem mentioned in https://github.com/HospitalRun/hospitalrun-frontend/issues/485#issuecomment-249384870. That is, when saving a patient, update the sequence accordingly if the patient `friendlyId` is larger.

gnowoel:
@jkleinsc Thanks! You're correct. We don't lock the sequence any more, so it's required to also address #572 to prevent duplicates.

Issue #51, opened 2015-04-02
Add patient copay field


jkleinsc:
Need a field to track patient responsibility. In the Philippines this field is called Patient counterpart.

jkleinsc:
While bigger than this issue, when we implement this, we should add [ember-i18n](https://github.com/jamesarosen/ember-i18n) so that we can customize the label for the Philippines

jkleinsc:
#141 is the ticket for internationalization

tangollama:
@jkleinsc can we close this?

jkleinsc:
@tangollama I don't think we ever added this field

wolfieorama:
my proposed plan is to add a field to capture whatever amount the patient
contributes on top of what the other parties take responsibility of.
My thought is to have it as a  non-required field handling different currencies (number
type) -- an addon may be to handle different currencies

taras:
@wolfieorama your plan looks good. 

I would suggest that we break this into 2 parts:
1. add this copay number field
2. make it work properly for with currency indicator

For 1, we might want to use the code that was introduced in #668
For 2, we need to figure out how to handle currency across the entire app (we might need a localize currency format)

jkleinsc:
@wolfieorama the intention is that we add a configuration to set a system wide currency, so any monetary values put into the system would be in that currency.  I don't think HospitalRun should try to support multiple currencies in one location.

wolfieorama:
@jkleinsc @taras I will jump on another issue as that is sorted

jkleinsc:
I am closing this issue as this enhancement request is satisfied by the custom forms feature.

PR #668, opened 2016-09-24
Add sanitizing number input

### LINKED ISSUE #116: Create new number component
Right now our number validation allows for numbering like 123,456 but codewise we are expecting numbers without commas.  We should create a component that will let people enter a number with commas but will save the value without commas.

Either that or we need to change how validation works to eliminate commas but that is not as friendly

## TODO
- [ ] Create number component

matteddy:
Fixes #116

This PR creates a `number-input` which sanitizes text to remove characters that are not numbers or a period. For example, if you were to paste "$1,000.00" into a number input it would change the value of the input to "1000.00". 

With these changes it is now much easier to extend `em-input` and add any sort of sanitization desired in the future.   

Outstanding Issues: 
There are [many countries](https://en.wikipedia.org/wiki/Decimal_mark#Countries_using_Arabic_numerals_with_decimal_comma), which use a comma as a decimal mark. This feature has the potential to cause confusion by removing commas which are important to actual value of the number e.g. "$100,00" one hundred dollars becoming "$10000" ten thousand dollars. 

Country specific configuration seems like something that may be a frequent issue for an app of this nature. It may be worth spending some time thinking about how to add such a configuration to the system.


taras:
@jkleinsc can you pls review this so we can get it merged in? I'd like to use this code for #51

jkleinsc:
Thanks for the PR @matteddy! I'm going to merge it in.

PR #672, opened 2016-09-25
upgrade ember to 2.9 beta 2

billybonks:
Upgrades ember-cli to 2.9 beta 2


jkleinsc:
@billybonks I would prefer to wait until Ember versions are out of beta to upgrade them.

billybonks:
NP :)

PR #686, opened 2016-10-01
Fix #496 text cursor icon 

### LINKED ISSUE #496: Cursor on links appears as a text cursor on Firefox and MS Edge
**Expected behavior:**
The cursor indicates a clickable object (i.e. link).

**Actual behavior:**
The cursor appears to be a text cursor giving no indication that e.g. clicking a patient name leads to the patient record.

**OS and Browser:**
OS: Windows 10
Browsers:
- Firefox 46.0.1
- Edge

Eric-Arellano:
Fixes #496.

- Add new rule to CSS to have un-disabled Ember data actions display the "Pointer" cursor icon (the hand), rather than the text selector icon that browsers were defaulting to in issue #496.

hr_cursor-icon_screenshot

**Inspired by Ember docs**
I found this fix in Ember Docs, bottom of this [page](https://guides.emberjs.com/v2.8.0/templates/actions/)

**_Problem: false positives**_
Be warned this leads to some false positives, where the browser renders the pointer/hand cursor when really the area isn't clickable. 

Noticed here:
- In lists where the actionable buttons are disabled, such as the screenshot's "medication" list's first entry
- In the patients page, the entire top patient section shouldn't be clickable because you're already on the patients page. However, this patient section renders correctly in every other location (i.e., from other pages, clicking the section will take you to the patient page).

So, you're replacing the problem of always showing the text cursor icon instead of correct pointer/hand icon, but this isn't perfect. I'd argue this is better, but maybe open another issue?


Eric-Arellano:
@HospitalRun/core-maintainers @tangollama FYI this now passes the tests after rerunning it.

jkleinsc:
@Eric-Arellano thanks for your work on this PR.   I'm not sure what to think about the false positives as they seem to be just as confusing as the original problem.  I think a proper fix would be to add a class to action items so we can apply styles specifically instead of generically applying to all.  @jglovier what do you think of this change?

tangollama:
@jglovier @jkleinsc are we going to complete or close out this PR. Thx again @Eric-Arellano for the contribution.

jkleinsc:
@tangollama I am going to close this one.

PR #706, opened 2016-10-02
Fixes 572: validating uniqueness of a patient's friendly ID

gnowoel:
I added a custom validator for dynamically showing if the display ID of a patient has already been taken. This should work when either adding or editing a patient. For the former case, we make sure the display ID is not used, and for the latter, we instead check if the display ID is only used by the current record.

screen shot 2016-10-02 at 11 46 38 pm

This is aiming to solve issue #572 partially. The remaining problem would be also validating the display ID on save, so that two people adding records simultaneously won't result in the same ID. I'll try to tackle that later as a separate task.

I added a custom validator because it seems `ember-validations` does not allow promises in an inline one.

Please correct me!


gnowoel:
If two users are adding patients at the same time, both records would get the same friendly ID. To prevent duplicates, I added a new [commit](https://github.com/HospitalRun/hospitalrun-frontend/pull/706/commits/a76002180c5eb9fd5fa1530c527f3e79b7e0c106) to force revalidating before saving a patient. So, if the ID has been taken, we'll see an error message in a modal dialog.

screen shot 2016-10-05 at 9 24 37 pm

Some rare race conditions might still occur because the enforcement happens in the application level, not the database. But considering the friend ID is not something we use to uniquely identify the records, perhaps the solution is good enough now.

What do you guys think?

jkleinsc:
@gnowoel thanks for the PR.  I have been thinking about this a bit and I think we need to go in a slightly different direction.  Here is what I am thinking:
1) The patient id should not be an editable field.  This reduces the complexity by not allowing users to pick an already used id and we do not need to check the id on updates.
2) When saving a new record, in the beforeUpdate we need to validate if the id has already been used.  If it has, grab the next available sequenced id and save the record with that.

cc @jglovier @tangollama

gnowoel:
Sounds good to me :)

gnowoel:
@jkleinsc Just added a new [commit](https://github.com/HospitalRun/hospitalrun-frontend/pull/706/commits/e5aa39390844666abdefc02ef748c399ba86d366). When saving a patient, now it will automatically use the next available one if the suggested friendly ID has been taken. The form field is not disabled, because I'm not sure about the UI part. Hope I'm at least in the correct direction.

app/patients/edit/controller.js
```diff
@@ -547,6 +548,14 @@ export default AbstractEditController.extend(BloodTypes, ReturnTo, UserSession,
     });
   },
 
+  beforeUpdate: function() {
```
jkleinsc: @gnowoel instead of validating through the model, you should just move the uniqueness code here.  Also, it should only run when the model is new (eg `if (this.get('model.isNew')`

The problem with the code the way it is right now if the model is invalid for other reasons (say for example the user didn't enter a first name, it would change the id).

app/models/patient.js
```diff
@@ -90,7 +90,8 @@ export default AbstractModel.extend(DOBDays, PatientName, {
       }
     },
     friendlyId: {
-      presence: true
+      presence: true,
+      uniqueness: true
```
jkleinsc: Remove the validation from the model and just do it in the controller instead.

app/validators/uniqueness.js
```diff
@@ -0,0 +1,45 @@
...
+    const database = this.get('database');
+    const id = this.get(this.property);
+    const maxValue = this.get('maxValue');
+    const query = {
+      startkey: [id, null],
+      endkey: [id, maxValue]
+    };
+
+    this.set('isValidating', true);
+    database.queryMainDB(query, 'patient_by_display_id')
```
jkleinsc: Move this code to the controller.

gnowoel:
@jkleinsc Thanks for your guidance. I [updated the code](https://github.com/HospitalRun/hospitalrun-frontend/pull/706/commits/8636809e5f5696adf09922ce2568885ef6594a01) again. Please correct me.

jkleinsc:
@gnowoel I think this PR is complete if you make one more change.  I mentioned disabling the friendly id from editing, but I realized all we need to do is remove the field from the form completely because it is already displayed in the header.  

So, to complete this PR can you remove this line?
https://github.com/HospitalRun/hospitalrun-frontend/blob/master/app/patients/edit/template.hbs#L132
`{{em-input property="friendlyId" label=(t 'labels.id') class="form-input-group required test-id"}}`

gnowoel:
@jkleinsc Thanks for your precise instruction. I have removed the ID field completely from the patient form in [this follow-up commit](https://github.com/HospitalRun/hospitalrun-frontend/pull/706/commits/e5274c99e456b6cc9032ea98a364cdef039fd150).

jkleinsc:
@gnowoel thanks for finishing up this PR!  Looks good to me, so I'll merge it in.

PR #710, opened 2016-10-04
fix size text menu left

rafaelmilanibarbosa:
Fixes #5314cdb.
- Change text size menu left


jkleinsc:
@rafaelmilanibarbosa What is the reason for this change?  "Fixes #5314cdb" isn't a valid issue.  Can you explain further what you are attempting to do here?

PR #715, opened 2016-10-07
Add Size Query Param for Elastic Search

### LINKED ISSUE #453: When using Elastic Search, only the first 10 patients are returned
Expectation is that the entire result set is returned, OR a message to the user letting them know how many were returned, out of the total set

Currently, only the first 10 are being returned (not sure what order)

Using the beta site, search for Ralph.  Locate Ralph Glick id = 0063.   Search again for Glick.   That record is not returned.    @jkleinsc did a first pass eval on the issue.

MacOX, Chrome

donaldwasserman:
Fixes #453.

Added configurable Elastic Search size option. 

I don't have elastic search running locally and it doesn't seem to be an issue with the server routes. Can submit a PR on those if you think it's necessary.

This is an exciting project


app/adapters/application.js
```diff
@@ -138,6 +140,11 @@ export default Adapter.extend(PouchAdapterUtils, {
         break;
       }
     }
+
+    if (!query.options.size) {
```
jkleinsc: @donaldwasserman I just tested this and I get the following error:
`vendor.js:40820 Error while processing route: patients.search Cannot read property 'size' of undefined TypeError: Cannot read property 'size' of undefined`
I believe this is because query.options is not passed in.

server/config-example.js
```diff
@@ -34,4 +34,4 @@ config.couchDbURL = config.getProtocol(config.couchDbUseSsl) + config.couchDbSer
 config.couchAuthDbURL = config.getProtocol(config.couchDbUseSsl) + config.couchCredentials() + config.couchDbServer + ':' + config.couchDbPort;
 // config.searchURL = 'http://localhost:9200'; ELASTIC SEARCH URL (OPTIONAL)
 config.serverInfo = 'Development Ember CLI server';
-module.exports = config;
+module.exports = config;
```
jkleinsc: This change shouldn't be here.

jkleinsc: Also, I realized that this size parameter never makes it to the call to elasticsearch.

donaldwasserman:
That explains why I couldn't get the server route to have an error. I'll
take a look at it this evening.

server/config-example.js
donaldwasserman: After stopping and thinking it through, it seems like ` _executeContainsSearch` is the only method that makes elastic search requests? I know that's a pretty basic question, but this is  pretty abstracted out for my simple brain.

jkleinsc: @donaldwasserman correct... _executeContainsSearch is what ends up calling to /search/hrdb/ which is a proxy to ElasticSearch (or a temp view search using couchdb).  

This part of the code needs to have the size parameter passed to it:
```javascript
Ember.$.ajax(searchUrl, {
  dataType: 'json',
  data: {
    q: queryString,
    size: variableWithSize //Replace variableWithSize with whatever variable has the size.
  },
  success: successFn
});
```

Also, even if you don't have ElasticSearch installed you can still test to see if the value is being passed along by running a search from Patient and looking in the developer console to see what url the app is passing along to the search proxy.  The url should end with ```&size=25```

donaldwasserman:
should be all set

app/adapters/application.js
```diff
@@ -57,10 +59,16 @@ export default Adapter.extend(PouchAdapterUtils, {
...
+        if (Ember.isEmpty(query.size)) {
+          query.size = this.get('_esDefaultSize');
+        }
+
         Ember.$.ajax(searchUrl, {
           dataType: 'json',
           data: {
-            q: queryString
+            q: queryString,
+            size: self.get('_esDefaultSize')
```
gnowoel: Didn't try the code, but it appears `this` should be used instead of `self`.

donaldwasserman:
I forgot to commit the code that changed this to self. should be good now.

jkleinsc:
Thanks for the PR!  Looks good to me, so I'll merge it in.

PR #724, opened 2016-10-12
Replace jshint with eslint. Closes #596 and closes #601.

### LINKED ISSUE #596: Linters

btecu:
`jshint` has merged with `eslint` and only `eslint` will be supported.

Note: Using `ember-cli-eslint` version `2.0.0` because the latest `3.0.0` has dropped support for node `0.10` and `0.12`.

Will follow additional PR to fix some newly discovered warnings.


btecu:
Note that `eslint-plugin-ember-suave` is included but not turned on.
I plan on submitting another pull request where I turn it on and fix all the errors.

jkleinsc:
@btecu we don't need to support node 0.10 and 0.12, so I would recommend using the latest ember-cli-eslint

btecu:
@jkleinsc Great, I've updated `ember-cli-eslint` to version `3.0.0`.

package.json
```diff
@@ -73,6 +73,7 @@
     "ember-suave": "4.0.1",
```
btecu: Good catch. Updated.

PR #738, opened 2016-10-13
Fix the `sort` list function in couchdb design docs

### LINKED ISSUE #45: Patient listing paging back button doesn't always go to start
There seems to be an issue where paging on patient listing doesn't let you go all of the way back to the first page.  Not sure what steps to reproduce are.

gnowoel:
Fixes #45.

Seems PouchDB does not recognize the `sortDesc` query parameter correctly. So I replaced it with `descending` as per [this doc](https://pouchdb.com/2014/04/14/pagination-strategies-with-pouchdb.html). Please correct me!

### DIFF
--- a/app/adapters/application.js
+++ b/app/adapters/application.js
@@ -152,8 +152,8 @@ export default Adapter.extend(PouchAdapterUtils, {
       if (query.options) {
         queryParams = Ember.copy(query.options);
         if (query.sortKey || query.filterBy) {
-          if (query.sortDesc) {
-            queryParams.sortDesc = query.sortDesc;
+          if (query.descending) {
+            queryParams.descending = query.descending;
           }
           if (query.sortKey) {
             queryParams.sortKey = query.sortKey;
--- a/app/adapters/application.js
+++ b/app/adapters/application.js
@@ -152,8 +152,8 @@ export default Adapter.extend(PouchAdapterUtils, {
       if (query.options) {
         queryParams = Ember.copy(query.options);
         if (query.sortKey || query.filterBy) {
-          if (query.descending) {
-            queryParams.descending = query.descending;
+          if (query.sortDesc) {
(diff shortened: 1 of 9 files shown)


jkleinsc:
@gnowoel thanks for the PR.  Unfortunately this PR causes the sorts to always be sorted ascending and this is because the "sortDesc" parameter is a special parameter used in our list functions (http://docs.couchdb.org/en/1.6.1/couchapp/ddocs.html#listfun) which is how we sort our views.  Making the change to pass along "descending" causes the following to happen:
1) The view gets sorted in descending order by the key for the via (in the case of patient it is sorted descending by id)
2) The list function reverses the order of the results.

The TLDR is that the descending parameter doesn't help us because that parameter sorts on the views key, not on the order parameter passed.  In order to do sorting on the order parameter, we have to pass it along to a list function which will reverse the order if sortDesc=true is passed.

jkleinsc:
@gnowoel I misspoke a bit above... The reason I was seeing the patients always sorted in ascending order is that the views didn't get updated to recognize the changing of the parameter from sortDesc to descending.  However, my point is that passing along descending=true only affects the sort order of the keys (see http://docs.couchdb.org/en/1.6.1/api/ddoc/views.html), not the columns we are trying to sort by.  the list functions are using sortDesc to determine sortOrder for a column.

gnowoel:
@jkleinsc Thanks for your detailed explanation. It really helps for me to understand the problems. I'll try to figure out a real solution.

gnowoel:
@jkleinsc PR updated. Hope this one would work :)

gnowoel:
@jkleinsc Yes, I missed those two views. Thanks for pointing out. PR updated.

jkleinsc:
@gnowoel thanks for updating! Everything looks good so I'll merge it in.  Thanks again!

PR #747, opened 2016-10-17
commented on the require function

cflatts:
I commented on the require function for Hacktoberfest.


jkleinsc:
@cflatts thanks for the PR!  I am going to close this PR because it doesn't necessarily help to have that comment there.

PR #656, opened 2016-09-19
making the cancel update button always visible

ShadabQ:
Fixes #569 make the save update button always visible.

**Changes proposed in this pull request:**
- make the footer-panel on /app/templates/components/edit-panel.hbs on the top.
- make its style as position:fixed.

cc @HospitalRun/core-maintainers @mnorbeck @tangollama @jglovier 
********************************************************************************************
<div class="panel panel-primary">
  
  <div class="panel-footer" style="position:fixed;">
    {{#unless editPanelProps.hideCancelButton}}
      {{#if editPanelProps.cancelButtonText}}
        <button class="btn btn-default warning" {{action 'cancel'}}>{{editPanelProps.cancelButtonText}}</button>
      {{else}}
        <button class="btn btn-default warning" {{action 'cancel'}}>{{t 'buttons.cancel'}}</button>
      {{/if}}
    {{/unless}}
    {{#if editPanelProps.showUpdateButton}}
      {{#if editPanelProps.disabledAction}}
        <button class="btn btn-primary on-white disabled-btn" {{action 'disabledAction'}}>{{editPanelProps.updateButtonText}}</button>
      {{else}}
        <button class="btn btn-primary on-white" {{action 'updateButtonAction'}} disabled={{editPanelProps.isUpdateDisabled}}>
          {{editPanelProps.updateButtonText}}
        </button>
      {{/if}}
    {{/if}}
    {{#each editPanelProps.additionalButtons as |button|}}
      <button class="{{button.class}}" {{action 'fireButtonAction' button.buttonAction}}>
        {{#if button.buttonIcon}}
          <span class="{{button.buttonIcon}}"></span>
        {{/if}}
        {{button.buttonText}}
      </button>
    {{/each}}
  </div>
  <div class="panel-body">
    {{yield}}
  </div>
</div>

PR #755, opened 2016-10-20
Address #569 - Contextual calls to action

### LINKED ISSUE #569: Make the Save (Add, Update) buttons always visible
The request is that any button to Save (Add or Update) would always be visible. Users would not need to scroll to the bottom of a page to find the button.    This could mean placing a second copy of the button at the top of the screen.   Request is to make this a standard across Hospital Run

James1x0:
I took a jab at #569. The difficult part is getting a nested controller's action to the parent controllers "section" context. I ended up just binding a context and dynamically setting additionalButtons. Lmk what you think; I'll be happy to continue iterations if needed. I've attached a screenshot.

Fixes #569.

- Removes `new patient` button from `patients.edit` route
- Adds `cancel`, `update` buttons to `patients.edit` route

/cc @HospitalRun/core-maintainers

screenshot at oct 19 21-14-22

### DIFF
--- a/app/patients/edit/route.js
+++ b/app/patients/edit/route.js
@@ -50,6 +50,18 @@ export default AbstractEditRoute.extend(PatientId, PatientVisits, PouchDbMixin, 
 
     visitDeleted: function(model) {
       this.controller.send('visitDeleted', model);
+    },
+
+    willTransition () {
+      let patientsController = this.controllerFor('patients');
+      if ( patientsController ) {
+        patientsController.setProperties({
+          additionalButtons: this.get('previousAdditionalButtons'),
+          newButtonAction: this.get('previousNewButtonAction'),
+          newButtonText: this.get('previousNewButtonText')
+        });
+      }
+      this._super(...arguments);
     }
   },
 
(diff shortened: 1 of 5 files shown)


jkleinsc:
@James1x0 thanks for the PR!  It seems that with this PR you are working upstream against the current architecture and I am wondering/thinking what this issue really needs is a refactoring of the "section" as #569 is addressed for the app as a whole, not just patients.

James1x0:
@jkleinsc I'm thinking the issue kind of morphed into a different spec, see https://github.com/HospitalRun/hospitalrun-frontend/pull/656#issuecomment-252362375

But I wouldn't have anything against attempting a slight refactor of the section to account for this.

jkleinsc:
@James1x0 I believe even with that comment the issue is looking for a change across the app or maybe I misunderstand your comment?

James1x0:
@jkleinsc I'm probably reading it at face value. Summarizing, @jglovier suggested we replace the button up top in patients and then an overall style adjustment would follow. I'm making the change utilizing the section APIs. Maybe there should be a WIP PR? Thoughts?

jkleinsc:
@James1x0 -- Yeah... the more I think about it, I think the issue needs clearer requirements/architectural direction before a PR.

jkleinsc:
@James1x0 I am closing this issue because I've come to the conclusion that this issue needs further definition/architectural oversight before a solution can be applied.  See this comment for more details: https://github.com/HospitalRun/hospitalrun-frontend/issues/569#issuecomment-255200000

PR #785, opened 2016-11-01
add acceptance test for patients.admitted

### LINKED ISSUE #699: Build acceptance test for patients.admitted
In order to effectively use the no missing translations test, we need to make sure that every page is visited during acceptance tests.

This issue tracks the missing `patients.admitted` acceptance test

for ember new commers the value `some.route` is the route that's missing you can find them under router.js

ZachEddy:
Fixes #699.

**Adds an acceptance test for `patients.admitted`**

Because `patients` and `patients.admitted` use the same partial, they can be tested identically.

To avoid duplicate code, I wrapped the `patients` and `patients.admitted` test inside a helper function named `testPatientList`. 

The logic for both tests is identical and unchanged, so it may not be worth testing these separately. I was just looking for a bite-size pull request to get oriented.

I'll start tackling more of these acceptance test issues.


ZachEddy:
Thanks for the reply, @jkleinsc! Your feedback makes a lot of sense. I'll start writing a test to meet the requirements you listed.

Issue #732, opened 2016-10-12
Cleanup based on `ember-suave` rules


btecu:
Continuation of #724.

**Plan**
Currently, all failing rules are turned off.
The plan is to go and fix all of the errors that the core maintainers agree on and submit those in individual (smaller) pull requests for faster merging and to avoid conflicts.

**Changes proposed**
- [x] array-bracket-spacing: `[ 1, 2 ]` -> `[1, 2]` #733
- [ ] camelcase: `some_ember_function` -> `someEmberFunction`
- [x] comma-spacing: `thing(a,b)` -> `thing(a, b)` #734
- [x] ember-suave/no-const-outside-module-scope: `const` -> `let` (inside modules) #735
- [ ] ember-suave/no-direct-property-access: `Ember.isNone()` -> `const { isNone } = Ember; isNone()` @amyrlam working on it
- [x] ember-suave/prefer-destructuring: `let brand = car.brand;` -> `let { brand } = car;` #769
- [ ] ember-suave/require-access-in-comments: most likely n/a
- [x] indent: consistent indentation (2 spaces) #772
- [x] key-spacing: `{ age:21 }` -> `{ age: 21 }` #741
- [x] keyword-spacing: `}else {` -> `} else {` #771
- [x] new-cap: `new report()` -> `new Report()` #810
- [ ] no-console: don't console log - questionable
- [x] object-curly-spacing: `{'foo': 'bar'}` -> `{ 'foo': 'bar' }` #770
- [x] object-shorthand: `doWork: function()` -> `doWork()` #809
- [x] one-var: `let a, b;` -> `let a; let b;` #736
- [x] operator-linebreak: `+` should be on the new line instead of end of previous line #773
- [x] prefer-spread: `obj.foo.apply(obj, args)` -> `foo.apply(obj, args)` #808
- [x] prefer-template `'My name is ' + name` -> ``My name is ${name}`` #745
- [x] quotes: `"` -> `'` #746
- [x] spaced-comment -> `//Comment` -> `// Comment`#739
- [x] space-infix-ops: `a+b` -> `a + b` #740

Will leave `no-console` off.

cc @HospitalRun/core-maintainers

jkleinsc:
Thanks @btecu!

amyrlam:
Hi, I'll take this one for now if that's ok?

> ember-suave/no-direct-property-access: Ember.isNone() -> const { isNone } = Ember; isNone()

[...]

PR #810, opened 2016-11-17
[Cleanup] eslint `new-cap`

btecu:
`new report()` -> `new Report()`

Part of #732.

### DIFF
--- a/app/inventory/barcode/controller.js
+++ b/app/inventory/barcode/controller.js
@@ -5,6 +5,8 @@ export default Ember.Controller.extend({
   barcodeUri: function() {
     let id = this.get('model.id');
     let name = this.get('model.name');
+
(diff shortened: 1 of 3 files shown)


app/inventory/barcode/controller.js
```diff
@@ -5,6 +5,8 @@ export default Ember.Controller.extend({
   barcodeUri: function() {
     let id = this.get('model.id');
     let name = this.get('model.name');
+
+    /* eslint new-cap: ['error', { 'capIsNew': false }]  */
```
jkleinsc: @btecu why is this override needed here?  I'm not seeing a `new` being used here.

jkleinsc: nevermind... I see why now.

PR #820, opened 2016-11-21
Trigger database syncing with Background Sync

### LINKED ISSUE #731: Syncing for offline use not working
**Expected behavior:**
When a user is online records should be synced to the user's browser for offline use.
**Actual behavior:**
Records are not synced for offline use.
**Steps to reproduce:**
1. Make sure you don't have offline records already synced.  Make sure the app isn't running in any tabs and then you can clear out any offline data stored by going to chrome://settings/cookies in a Chrome tab.  You should see a screen like this:
screen shot 2016-10-12 at 4 14 59 pm
Type in localhost (or whatever host your app is running on) and you should see an entry like the one in the screenshot.  Click on the entry and then click on the `X` on the right hand side of the screen.
2. Make sure the app is running via `ember serve` or via hospitalrun-server.
3. Navigate to the application and verify that records are available.  For testing simplicity I would use the sample-data set here: 
https://github.com/HospitalRun/hospitalrun-frontend/blob/master/sample-data.txt
4. Leave the app up for 5 minutes to make sure syncing has time to complete.
5. Turn off the server and then check in the app if the records you saw while the server was running are still there.
N/A
**OS and Browser:**
Chrome 53.0.2785.143 (64-bit)

gnowoel:
Fixes #731.

Do initial sync in the main thread. Subsequent syncing is triggered with a Background Sync event and only happens in the Service Worker.

### DIFF
--- a/app/adapters/application.js
+++ b/app/adapters/application.js
@@ -9,6 +9,60 @@ const {
   }
 } = Ember;
 
+const { localMainDB, remoteDB } = window;
+
+function foregroundSync() {
+  console.log('foreground syncing');
+  return localMainDB.sync(remoteDB).on('error', function(err) {
+    console.log('foreground sync error:', err);
+  });
+}
+
+function backgroundSync() {
+  console.log('registering background sync');
+  return navigator.serviceWorker.ready
+    .then(function(reg) {
+      return reg.sync.register('sync');
+    })
+    .then(function() {
+      console.log('background sync registration succeeded');
+    })
+    .catch(function() {
+      console.log('background sync registration failed');
+    });
+}
+
+function watchLocalDB() {
+  console.log('watching local DB');
+  return localMainDB.changes({
+    since: 'now',
+    live: true
+  }).on('change', function(change) {
+    console.log('local change: ', change);
+    return backgroundSync();
+  }).on('error', function(err) {
+    console.error('local err: ', err);
+  });
+}
+
+function watchRemoteDB() {
+  console.log('watching remote DB');
+  return remoteDB.changes({
+    since: 'now',
+    live: true
+  }).on('change', function(change) {
+    console.log('remote change: ', change);
+    return backgroundSync();
+  }).on('error', function(err) {
+    console.error('remote err: ', err);
+  });
+}
+
+foregroundSync().then(function() {
+  watchLocalDB();
+  watchRemoteDB();
+});
+
 export default Adapter.extend(PouchAdapterUtils, {
   database: Ember.inject.service(),
   db: Ember.computed.reads('database.mainDB'),
--- a/app/adapters/application.js
+++ b/app/adapters/application.js
@@ -58,10 +58,12 @@ function watchRemoteDB() {
   });
 }
 
-foregroundSync().then(function() {
-  watchLocalDB();
-  watchRemoteDB();
-});
+if (localMainDB && remoteDB) {
+  foregroundSync().then(function() {
+    watchLocalDB();
+    watchRemoteDB();
+  });
+}
 
 export default Adapter.extend(PouchAdapterUtils, {
   database: Ember.inject.service(),
--- a/app/services/database.js
+++ b/app/services/database.js
@@ -44,12 +44,15 @@ export default Ember.Service.extend(PouchAdapterUtils, {
       }
       let url = `${document.location.protocol}//${document.location.host}/db/main`;
 
-      this._createRemoteDB(url, pouchOptions)
+      let localMainDB = this._createLocalDB('localMainDB', pouchOptions);
+      let remoteDB = this._createRemoteDB(url, pouchOptions);
+
+      remoteDB
       .catch((err) => {
         if ((err.status && err.status === 401) || configs.config_disable_offline_sync === true) {
           reject(err);
         } else {
-          return this._createLocalDB('localMainDB', pouchOptions);
+          return localMainDB;
         }
       }).then((db) => resolve(db))
       .catch((err) => reject(err));
@@ -162,6 +165,7 @@ export default Ember.Service.extend(PouchAdapterUtils, {
   _createRemoteDB(remoteUrl, pouchOptions) {
     return new Ember.RSVP.Promise(function(resolve, reject) {
       let remoteDB = new PouchDB(remoteUrl, pouchOptions);
+      window.remoteDB = remoteDB;
       // remote db lazy created, check if db created correctly
       remoteDB.info().then(()=> {
         createPouchViews(remoteDB);
@@ -176,6 +180,7 @@ export default Ember.Service.extend(PouchAdapterUtils, {
   _createLocalDB(localDBName, pouchOptions) {
     return new Ember.RSVP.Promise(function(resolve, reject) {
       let localDB = new PouchDB(localDBName, pouchOptions);
+      window.localMainDB = localDB;
       localDB.info().then(() => {
         createPouchViews(localDB);
         resolve(localDB);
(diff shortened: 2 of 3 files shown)


jkleinsc:
@gnowoel thanks for your work on this PR.  I think we need to hold off on further work on this until this PR for PouchDB is completed: https://github.com/pouchdb/pouchdb/pull/5903.  I believe this maybe why you are having trouble doing the initial sync.

Basically, PouchDB wasn't testing fetch properly in their test suite, so it appears there may be some issues with using fetch with PouchDB, which is needed to use PouchDB within a Service Worker.

Also, in looking at your code, listening for changes in the Service Worker will not work because the service worker can be killed at any time by the browser, so we can't rely on the changes listener to be long running.  

I haven't had time to test the concept in full but I think the solution for HospitalRun should be as follows in browsers that support background sync and service workers

1. The UI thread just has a remote DB
2. The Service Worker intercepts all network requests to remote DB (the code in HR already does this).
3. On network requests that are updating data, register a background sync event if the network request fails.
4. Server side Push API is used to push changes from server side to local db.

gnowoel:
@jkleinsc My motivation of the PR was to fix the upfront bug, while laying ground work for possible future update. I totally respect your decision to hold it off, but I really think we could perhaps improve this PR first. Here're the reasons just for your consideration.

> ... listening for changes in the Service Worker will not work...

Actually, we listen for database changes in [the main thread](https://github.com/gnowoel/hospitalrun-frontend/blob/875f4978c3ba264f3039752d807cd99e0aef6825/app/adapters/application.js#L62-L63) (in the `application` adapter). The Server Worker on the other hand only responds to the `sync` event fired on database change.

> The UI thread just has a remote DB

We created a PouchDB instance in the UI thread. This is for doing the initial sync and watching database changes. Considering the initial sync in fact only happens once (because the data would be persisted in IndexedDB) and watching database changes is a relatively [lightweight](https://github.com/pouchdb/pouchdb/issues/5851#issuecomment-259197073) operation, the performance impact should be acceptable.

> On network requests that are updating data, register a background sync event if the network request fails.

I didn't do intensive testing, but this seems to be working already. If there're any bugs, I'd be happy to fix them.

> Server side Push API is used to push changes from server side to local db.

It's a great idea for sure. However, as I briefly mentioned above, while it's not ideal, watching database changes in the UI thread should be acceptable. So, it's more like an enhancement than a necessity for me. Could we leave the watching code in the UI thread for now, and worry about how to integrate Push API (which requires more work because it needs service-side cooperation) for later?

> I think we need to hold off on further work on this until this PR for PouchDB is completed...

If the bug is fixed, maybe we can move the initial syncing code into the Service Worker. But it also sounds more like an enhancement to me. The initial sync only happens *initially*, when we start the app in the browser for the very first time. Once the databases has been fully synced for once, all subsequent calling to the [`one-shot sync function`](https://github.com/gnowoel/hospitalrun-frontend/blob/875f4978c3ba264f3039752d807cd99e0aef6825/app/adapters/application.js#L14-L19) would do about nothing.

**Suggesion**

While I totally respect your decision to hold it off, I would suggest that we fix the bug first. If we stop here, we get something that works already. If we want to improve it later by introducing advanced techniques like Pushing API, the current implementation shouldn't get in the way.

Please correct me if I'm completely wrong.

jkleinsc:
@gnowoel I misread your code.  I mistakenly thought the changes listeners were in the service worker.  Let me take a closer look at your code and what you are proposing.

Can you check on one thing on your end?  It looks like Travis CI is failing and I suspect if you run `ember test` locally you will see the errors as well.  The full output from Travis is available here:
https://travis-ci.org/HospitalRun/hospitalrun-frontend/builds/177716747

The error I am seeing is:
'Error: Failed to create an instance of \'adapter:application\'. Most likely an improperly defined class or an invalid module export. at http://localhost:7357/assets/vendor.js, line 12015\n' }

gnowoel:
@jkleinsc I missed the test part. Sorry about that.

The failure in the CI is due to undefined databases. We created databases when initializing the `database` service and I guess this means it's triggered by the first page request. However, there won't be a page request when running the app in a CI.

I updated the code to just skip the initial sync unless database are created.

By the way, do you think it's a good idea to implement a poor man's Periodic Sync with `setInterval()`, which repeatedly triggers lightweight Background Sync events from the UI thread?

app/services/database.js
```diff
@@ -162,6 +165,7 @@ export default Ember.Service.extend(PouchAdapterUtils, {
   _createRemoteDB(remoteUrl, pouchOptions) {
     return new Ember.RSVP.Promise(function(resolve, reject) {
       let remoteDB = new PouchDB(remoteUrl, pouchOptions);
+      window.remoteDB = remoteDB;
```
jkleinsc: Using a global is unneeded.  The application adapter has access to the database service, so it can use that to get the remoteDb via:
```js
this.get('database').get('mainDB')
```

app/adapters/application.js
```diff
@@ -9,6 +9,62 @@ const {
   }
 } = Ember;
 
+const { localMainDB, remoteDB } = window;
```
jkleinsc: These values should be pulled from the database service.  For example:
```js
let remoteDB = this.get('database').get('mainDB');
```

app/services/database.js
```diff
@@ -176,6 +180,7 @@ export default Ember.Service.extend(PouchAdapterUtils, {
   _createLocalDB(localDBName, pouchOptions) {
     return new Ember.RSVP.Promise(function(resolve, reject) {
       let localDB = new PouchDB(localDBName, pouchOptions);
+      window.localMainDB = localDB;
```
jkleinsc: Like remoteDB, Using a global is unneeded, but we should set the localDB as a property on the database service so that the application adapter has access to it:
```js
this.set('localDb', localDb);
```
Then in the application adapter you can access via:
```js
let localDB = this.get('database').get('localDB');
```

jkleinsc:
@gnowoel as far as `setInterval`, I think that is unnecessary for now.

tangollama:
@gnowoel, do you think you'll be willing / able to make the recommended changes from @jkleinsc?

gnowoel:
@tangollama Sorry for the long delay. I didn't forget it, but I was trapped by my own life's struggle. Your reminder reminds me that there's no better time than now. And yes, I'm willing to finish what I started, at least give it another try before it's too late.

jkleinsc:
@gnowoel My apologies here.  Sorry to hear about your own life's struggle - I hope everything is going ok.  @tangollama and I weren't in sync on this.  I have been doing some work on this whole issue and given that we are trying to get a 1.0 release out this month, I need to take primary responsibility for this issue, so please feel released from taking care of this.  I am going to close this PR, but please feel free to contribute in the future!

gnowoel:
@jkleinsc Thanks for your kind words. It means a lot to me!

I was trying to update PR today, based on your very inspiring comment above. It's only partially completed, but along the way, I've found a few things that might be useful.

* The local and remote PouchDB instances are only available after logging in, which makes it difficult to setup database syncing in the application adapter.
* The application adapter seems to be initialized for every request, but we only want to setup syncing once.
* The `db` of the application adapter would change based on the connection status, so it might be a challenge to correctly use `onChange()` to trigger a background sync.

I ended up defining the sync-related methods in a service, and then injecting it in the module that handles authentication. That way, I can setup syncing just by adding a `then()` block:

```js
authenticate(credentials).then(() => this.get('sync').setup());
```

I'm not sure if it's a good idea. Perhaps Ember initializers are the way to go. I don't know. But I'm sure your solution will be much better.

Thanks again!

(I'd like to attach my latest update below. It's just a prototype. So, please ignore it if it's useless.)

gnowoel:
Oops. The code is not showing. Guess it's because the issue is closed. Never mind :) In case you want to take a look, please see [here](https://github.com/gnowoel/hospitalrun-frontend/tree/patch-6).

PR #821, opened 2016-11-24
Removes unnecessary `resetNamespace`s

locks:
This is a code style improvement with no issue associated.

- Remove unnecessary `{ resetNamespace: true }` for top-level routes.

The routes declared directly beneath `Router.map` are already top-level, so the reset is unnecessary.
This has the added benefit of removing some visual cruft, because the few `resetNamespace`s left are actually meaningful, making visual parsing easier, as well as grepping.

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -11,7 +11,9 @@ Router.map(function() {
     this.route('address');
     this.route('loaddb');
     this.route('lookup', { path: '/' });
-    this.route('users', function() {
+    this.route('users', {
+      resetNamespace: true 
+    }, function() {
       this.route('edit', { path: '/edit/:user_id' });
     });
     this.route('roles');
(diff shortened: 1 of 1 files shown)


app/router.js
```diff
@@ -7,25 +7,19 @@ const Router = Ember.Router.extend({
...
-    resetNamespace: true
-  }, function() {
+  this.route('admin', function() {
     this.route('address');
     this.route('loaddb');
     this.route('lookup', { path: '/' });
-    this.route('users', {
-      resetNamespace: true
-    }, function() {
+    this.route('users', function() {
```
jkleinsc: This change is causing tests to fail because users sits at `/app/users`, not `/app/admin/users`

locks:
@jkleinsc oops, I think I misread the indentation! Do you want me to squash the commits as well?

jkleinsc:
@locks I can squish the commits via GithHub, so you don't need to worry about that.

jkleinsc:
Thanks @locks!  I'll merge it in.

locks:
Thanks :D

PR #840, opened 2016-12-08
Add model unit test for add-diagnosis

mkly:
- Add model unit test for add-diagnosis


mkly:
Another small one. Let me know if you would rather I group these together as I'm likely going to be pushing up some more model unit tests. :beer:

jkleinsc:
@mkly I think it is fine to submit individual PRs but if you want to group them together that is ok too.

PR #841, opened 2016-12-08
Add unit test for select-values

mkly:
- Add unit test for select-values

Test aligns with current behavior if non-array value is passed in as the
first argument.


mkly:
Feel like this behavior may want to change to throwing and Error that gets handled as simply returning `undefined` results in a broken state at that point.

PR #864, opened 2016-12-13
Fix math in InventoryPurchaseItem.costPerUnit and add math util for rounding

mkly:
Fixes math in InventoryPurchaseItem.costPerUnit and add math util for rounding

- Fix incorrect calculation `InventoryPurchaseItem.costPerUnit` to round
  property to the nearest 100th
- Add test for demostrating issue in `inv-purchase-test.js`
  "costPerUnit properly round"
- Add unit model test inv-purchase InventoryPurchaseItem
- Add math util with function `round100`
- Add tests for `round100` function


mkly:
Using `fixedTo(2)` fails on numbers like 35.555(rounds to 35.56). Normally this isn't a big deal with displays but since these are financials and I've seen various discussions about reporting I figure it would be wise to get a bit closer.

It is possible that this is not the only place this issue exists. I haven't looked for other spots as I wanted to make sure this was an acceptable solution going forward.

It may also be a better choice to instead bring in a library specifically designed to handler JavaScript's various math quirks. The advantage is that it would likely handle further edge cases than our internal implementations. The disadvantage is that is adds one more library to maintain and increases the overall download size of the application.

mkly:
Sounds good. Will do. Thanks

mkly:
Barring any CI issues this should be ready for review. Hopefully I understood you correctly in the use of the mixin here. I went ahead and added a little test coverage for NumberFormat while I was in there. Thanks again for taking the time for the review. Still learning my way around the code base and getting up to speed with the preferred conventions. :tropical_drink:

mkly:
Also, let me know if you want me to squash these commits down and open a new PR for merging

jkleinsc:
Looks good to me @mkly!  I can squash the commits here on GitHub.

PR #871, opened 2016-12-14
Define Incorrect Behavior of haveInvoiceItems

mkly:
- Define incorrect behavior of `InventoryBatch.haveInvoiceItems()` in
  model unit test
    
`haveInvoiceItems` returns `true` if `invoiceItems` is empty. This
should likely be reversed or method name should change.

### DIFF
--- a/app/models/inventory-batch.js
+++ b/app/models/inventory-batch.js
@@ -7,7 +7,7 @@ import Ember from 'ember';
 export default AbstractModel.extend({
   haveInvoiceItems() {
     let invoiceItems = this.get('invoiceItems');
-    return (Ember.isEmpty(invoiceItems));
+    return !Ember.isEmpty(invoiceItems);
   },
 
   validations: {
--- a/app/models/inventory-batch.js
+++ b/app/models/inventory-batch.js
@@ -16,7 +16,7 @@ export default AbstractModel.extend({
     },
     inventoryItemTypeAhead: {
       presence: {
-        if(object) {
+        unless(object) {
           return object.haveInvoiceItems();
         }
       }
@@ -27,7 +27,7 @@ export default AbstractModel.extend({
         messages: {
           greaterThan: 'must be greater than 0'
         },
-        if(object) {
+        unless(object) {
           return object.haveInvoiceItems();
         }
       }
@@ -38,7 +38,7 @@ export default AbstractModel.extend({
         messages: {
           greaterThan: 'must be greater than 0'
         },
-        if(object) {
+        unless(object) {
           return object.haveInvoiceItems();
         }
       }
--- a/tests/unit/models/inventory-batch-test.js
+++ b/tests/unit/models/inventory-batch-test.js
@@ -1,0 +1,22 @@
+import { moduleForModel, test } from 'ember-qunit';
+
+moduleForModel('inventory-batch', 'Unit | Model | inventory-batch', {
+  needs: [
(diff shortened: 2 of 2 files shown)


mkly:
I did not want to simply "fix" this as I was unsure if the behavior needed to be preserved. This also causes issues in the validations of the model as well. Let me know after merging this and I can fix the method and update the tests to match.

jkleinsc:
@mkly it looks like you came across a bug.  There is no reason that haveInvoiceItems should be operating the way that it is.  Can you correct haveInvoiceItems as part of this PR instead of adding an incorrect test?

mkly:
Barring any issues in CI this should be ready for review. I did some manual testing for the application but I was not fully able to understand where these validations are being used. You may want to have a quick look around to make sure this does not break something that was dependent on this returning the incorrect value or point me in the direction of by routes to test and I would be more than happy to test it further.

mkly:
Absolutely makes sense. This was part of my concern with changing the behavior. I'll make those adjustments. Thanks for testing it out.

app/models/inventory-batch.js
```diff
@@ -16,7 +16,7 @@ export default AbstractModel.extend({
     },
     inventoryItemTypeAhead: {
       presence: {
-        if(object) {
+        unless(object) {
```
jkleinsc: This doesn't resolve the issue.  ```unless(object)``` isn't the same as ```unless(object.haveInvoiceItems)```
The validation needs to fire when object.haveInvoiceItems is false.

mkly: It is very possible that I am missing something, but my understanding is that `unless` is the inverse of `if`. Meaning that `unless` will skip validation if the method returns `true` as opposed to `if` which will skip validation if the method return `false`. The pseudo code interpretation would be "Unless the object has invoice items, trigger validation".

An example of this is
https://github.com/mkly/hospitalrun-frontend/blob/af38a215463682ef4f31b6774078cb3145ae4497/tests/unit/models/inventory-batch-test.js#L34-L36
The validation is skipped because `invoiceItems` is non empty and therefore `haveInvoiceItems` returns `true`
The opposite is
https://github.com/mkly/hospitalrun-frontend/blob/af38a215463682ef4f31b6774078cb3145ae4497/tests/unit/models/inventory-batch-test.js#L32
The validation here is run because `invoiceItems` is not set(empty) and therefore `haveInvoiceItems` returns `true`. This is why the assertion of `testInvalidPropertyValues` passes here but not in the above linked test. Because in this test the validation did run.

I probably could have made this pull request easier to follow by just appending `!` in order to invert the call to `haveInvoiceItems`. I just thought using `unless` was a little clearer to someone coming into this fresh. That said, `unless` is not a conditional keyword in JavaScript like it is in ruby and perl. Let me know if I'm way off here and my apologies in advance if I am missing the obvious. And if you would prefer I can just invert with `!` and use `if` instead of `unless`.

Thanks again for taking the time to look this over.

jkleinsc: @mkly my bad -- I misread the code.... I translated it as a regular javascript expression, eg if object is defined, return the value from object.hasInvoiceItems() to decide if validation fires.

You were correct in using unless and I think it is fine to use it in the project.

mkly: Glad to hear I didn't mess something up. I worry with these pull requests I'll pull you away from more pressing matters if I start introducing bugs and other issues. Thanks as always.

PR #879, opened 2016-12-15
Unit test patient model mixin

mkly:
- Add mixin unit test for patient-name
- Add unit model test for patient


mkly:
A little of both. I'm just picking things off as I see them. I figure the code base can always benefit from a little more coverage and I'm using them to get up to speed to start helping with functional requirements and issues. In this case I probably just forgot. Leave it open and I'll add those since it looks pretty simple to add. Thanks again. :tropical_drink:

PR #881, opened 2016-12-15
Show Failing function in photo model (Test Intentionally Fails)

mkly:
- Add test that will fail with
  `this.getPatientDirectory is not a function`

`getPatientDirectory` does not appear to be defined anywhere and I do
not believe `downloadImageFromServer` is used anywhere as well.

Recommend removing the method or at minimum adding documentation that it
throws an error if used

### DIFF
--- a/tests/unit/models/photo-test.js
+++ b/tests/unit/models/photo-test.js
@@ -1,0 +1,23 @@
+import { moduleForModel, test } from 'ember-qunit';
+import Ember from 'ember';
+
+moduleForModel('photo', 'Unit | Model | photo', {
+  needs: [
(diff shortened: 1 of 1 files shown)


mkly:
This PR should fail in CI since the function itself cannot complete without error. Not sure what the preferred route here is. Let me know if we want to remove it or simply document that it does not currently work.

jkleinsc:
@mkly I took a look at that function and I think it was a carry over from a previous project and it isn't needed for HospitalRun.  It should be removed.

mkly:
Sent PR #887 that will close this PR upon merge.

PR #913, opened 2017-01-05
Modified medication model

hsorellana:
Re-write of the medication model, using the ember style guides from Dockyard


app/models/medication.js
```diff
@@ -109,29 +102,22 @@ export default AbstractModel.extend(CanEditRequested, DateFormat, MedicationDeta
...
             quantityToCompare = object.get('inventoryItem.quantity');
           }
-          if (requestQuantity > quantityToCompare) {
-            // force validation to fail
-            return true;
-          } else {
-            // There is enough quantity on hand.
-            return false;
-          }
+          return requestQuantity > quantityToCompare;
```
jkleinsc: While this technically does the same thing as before, the comments were intended to help others understand what is going on here, so I would prefer to leave them in.

tangollama:
@hsorellana could you re-include that comment or let me know that we should close the PR. Thx either way.

jkleinsc:
Closing in favor of #1065

PR #919, opened 2017-01-05
Modified visit model

hsorellana:
Re-write of the `visit` model, using the ember style guides from Dockyard


app/models/visit.js
```diff
@@ -36,77 +36,75 @@ function paymentStateAcceptance(object) {
...
     return diagnosisList;
   }),
 
   hasAppointmentLabel: computed('hasAppointment', function() {
-    let hasAppointment = this.get('hasAppointment');
-    let i18n = this.get('i18n');
-    if (hasAppointment === true) {
+    let hasAppointment = get(this, 'hasAppointment');
+    let i18n = get(this, 'i18n');
+    if (hasAppointment) {
```
jkleinsc: This line should remain as it was before because 
```js
if (hasAppointment === true) {
```
Only returns true if and only if hasAppointment has a value of true
```js
if (hasAppointment) {
```
Returns true even if hasAppointment has a value other than true (eg hasAppointment = 'N' would return true)

tangollama:
@hsorellana same here. I'd be glad to include this PR if you can respond to the requested change. Thx.

hsorellana:
@tangollama i'll work on this. if i have any trouble, i'll mention it here ;)

hsorellana:
Hey guys, i have a problem here. The thing is i deleted my fork of HR, and formatted my laptop, so is there a way to clone this pull request locally?? @tangollama

jkleinsc:
@hsorellana you can access your file here: https://github.com/hsorellana/hospitalrun-frontend/blob/524fe3bb17c0a675a861136500013148e059b7e4/app/models/visit.js

jkleinsc:
Closing in favor of #1064

PR #945, opened 2017-02-17
User role homescreen

tangollama:
@jkleinsc this is for review, not necessarily for merging. The idea here is to overcome the feature issue (and therefore usability concern) of users logging into a back “what do you want to do?” screen. Let’s discuss.


tangollama:
@jkleinsc can you review this PR for the purposes of getting into 1.0?

app/mixins/setup-user-role.js
```diff
@@ -3,6 +3,7 @@ export default Ember.Mixin.create({
   setupUserRole() {
     let session = this.get('session');
     let userRole = session.get('data.authenticated.role');
+    session.set('data.authenticated.userRole', userRole);
```
jkleinsc: Why not just pull the user's role from data.authenticated.role?  Why double store it?

app/mixins/navigation.js
```diff
@@ -341,5 +341,20 @@ export default Ember.Mixin.create({
 
       return nav;
     });
-  })
+  }),
+
+  findNavItemByRoute: function(route) {
+    return this.navItems.find(function(el) {
```
jkleinsc: I don't think you want to use find here.  According to the docs 
[find](http://emberjs.com/api/classes/Ember.Array.html#method_find) "Returns the first item in the array for which the callback returns true".  You are not returning true anywhere here and the way this code is written, if the item is found in a subnav item, it is going to return the parent item, not the subnav item.  You could accomplish what you need with forEach loops.

app/routes/index.js
```diff
@@ -1,6 +1,34 @@
...
-export default Ember.Route.extend(AuthenticatedRouteMixin, {
+const { inject } = Ember;
+
+export default Ember.Route.extend(AuthenticatedRouteMixin, Navigation, SetupUserRole, UserRoles, {
+  session: inject.service(),
+  beforeModel() {
+    // this.setupUserRole();
+    let session = this.get('session');
+    if (session != null) {
+      let role = session.get('data.authenticated.userRole');
```
jkleinsc: This should pull data.authenticated.role not data.authenticated.userRole, particularly if you want the acceptance test I added to pass.  😄

tangollama:
@jkleinsc I think I've satisfied the requested changes. Happy to discuss the why's Monday.

tangollama:
@jkleinsc All 16 acceptance tests pass for this locally (http://localhost:4200/tests?filter=Testing%20User%20Role%20homescreen). Travis appears to be timing out with a few of these. What shall we do?

PR #958, opened 2017-02-25
Move modal alert to inline

### LINKED ISSUE #377: Show patient update confirmation as inline alert rather than modal
**Current experience:** When you update a patient record, you are shown a confirmation dialog as a modal that lets you go back, or simply close the dialog. It is superfluous and requires an extra click to keep working if you don't want to return to the patient record screen.

image

**Better experience:** Instead, we should render the alert inline just above the button so that you don't have the extra step of dismissing the modal if you want to do anything other than returning to the patient index. We can still include the patient index link in the confirmation alert for convenience if that is your next destination.

**Steps to reproduce:** Go to a patient record and update any piece of information.

melissahie:
Fixes #377 

- Take out modal
- Show inline alert above button
- Hide inline alert after 3000ms

image

@billybonks @jglovier

### DIFF
--- a/app/patients/edit/controller.js
+++ b/app/patients/edit/controller.js
@@ -645,13 +645,9 @@ export default AbstractEditController.extend(BloodTypes, DiagnosisActions, Retur
 
   afterUpdate(record) {
     this._updateSequence(record).then(() => {
-      this.send('openModal', 'dialog', Ember.Object.create({
-        title: this.get('i18n').t('patients.titles.savedPatient'),
-        message: this.get('i18n').t('patients.messages.savedPatient', record),
-        updateButtonAction: 'returnToPatient',
-        updateButtonText: this.get('i18n').t('patients.buttons.backToPatients'),
-        cancelButtonText: this.get('i18n').t('buttons.close')
-      }));
+      $('.message').show();
+      $('.message').text(this.get('i18n').t('patients.messages.savedPatient', record));
+      $(".message").delay(3000).fadeOut(100);
     });
   }
 
--- a/app/patients/edit/template.hbs
+++ b/app/patients/edit/template.hbs
@@ -464,4 +464,5 @@
       {{/unless}}
     </div>
   {{/em-form}}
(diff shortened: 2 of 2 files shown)


app/patients/edit/controller.js
```diff
@@ -645,13 +645,9 @@ export default AbstractEditController.extend(BloodTypes, DiagnosisActions, Retur
...
-      this.send('openModal', 'dialog', Ember.Object.create({
-        title: this.get('i18n').t('patients.titles.savedPatient'),
-        message: this.get('i18n').t('patients.messages.savedPatient', record),
-        updateButtonAction: 'returnToPatient',
-        updateButtonText: this.get('i18n').t('patients.buttons.backToPatients'),
-        cancelButtonText: this.get('i18n').t('buttons.close')
-      }));
+      $('.message').show();
+      $('.message').text(this.get('i18n').t('patients.messages.savedPatient', record));
+      $(".message").delay(3000).fadeOut(100);
```
jkleinsc: This line is erroring out in our tests because it should use single quotes instead of double quotes.

dasManaswini:
@jkleinsc Can I help you add inline alert to the edit-panel component and fix this travis error?

jkleinsc:
@dasManaswini this is @melissahie's PR so as far as fixing the Travis error, that would need to be done from @melissahie's fork and branch here: https://github.com/melissahie/hospitalrun-frontend/tree/enhancement/move_modal_to_inline, so you would need to fork that repo, fix the error and submit your own PR.

In regards to adding this to the edit-panel component, I would create a fork off of this repo (https://github.com/HospitalRun/hospitalrun-frontend) and modify https://github.com/HospitalRun/hospitalrun-frontend/blob/master/app/components/edit-panel.js and https://github.com/HospitalRun/hospitalrun-frontend/blob/master/app/templates/components/edit-panel.hbs to temporarily display the markup when a value is set on the edit-panel component.

Hope that helps.  Let me know if you have further questions.

dasManaswini:
I am still working on this edit- panel component. Please add a Work in progress label

PR #973, opened 2017-03-06
inline alert added to patient edit screen

### LINKED ISSUE #377: Show patient update confirmation as inline alert rather than modal

dasManaswini:
Fixes #377 
Progress in work of #958 

- Added inline alert above button instead of modal 
- Hide inline alert after 3000ms

one

cc @jglovier @jkleinsc @tangollama

### DIFF
--- a/app/patients/edit/controller.js
+++ b/app/patients/edit/controller.js
@@ -647,7 +647,7 @@ export default AbstractEditController.extend(BloodTypes, DiagnosisActions, Retur
     this._updateSequence(record).then(() => {
       $('.message').show();
       $('.message').text(this.get('i18n').t('patients.messages.savedPatient', record));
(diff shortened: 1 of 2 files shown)


jkleinsc:
Closing per @dasManaswini's request

Issue #976, opened 2017-03-07
Tests fail when ran between 23:00 and 00:00 hours (Eastern, -5:00)


sukima:
**Expected behavior:**

Tests to pass.

**Actual behavior:**

Some tests fail. Near the top of the list is appointments.

**Steps to reproduce:**

1.  Change your computer clock time to 23:40 (11:40 PM) Eastern time zone
2.  Run tests in a browser (`http://localhost:4200/tests`)

**Screenshots (if applicable):**

<img width="814" alt="screen shot 2017-03-07 at 23 57 40" src="https://cloud.githubusercontent.com/assets/70075/23657554/b9483a8a-030c-11e7-9399-10bfa50a4562.png">

**OS and Browser:**

`Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/56.0.2924.87 Safari/537.36`

mkly:
I was able to recreate this by setting the global time in the acceptance test via `sinon.useFakeTimers()`. I haven't looked into the actual issue but hopefully the new failing tests may make it easier to resolve.

PR #978, opened 2017-03-07
Add failing tests for appointments after 23:00

mkly:
See #976 

- Add failing tests for appointments after 23:00

This sets the time to 23:40 so that the issues described in issue #976
are shown without setting the global system time

Does not resolve issue, only sets tests to properly fail for it


mkly:
@jkleinsc Those tests should have only applied to that one test section. I'm pushing a new commit to only apply to that or you may get additional false negatives with "Warning!!!". Sorry about that.

jkleinsc:
Thanks @mkly.  I am looking into this now.

PR #1001, opened 2017-03-19
OPD UX Changes

### LINKED ISSUE #996: OPD Visit Report - minor UX changes around view and print
This is an enhancement request.

1.  On the Visit page, on the reports tab.  Change the label of the button from "edit" to "view".   When you go to the report page, you really cannot edit anything, so "Edit" is misleading.   

2. On the Visit page, on the reports tab, next to the (now View) button, please add a Print button.   

3. On the OPD report page itself, remove the "update" button.   It is not really needed.   When you open the report, you are already getting an "update" or "refresh" of the report, based on the latest visit info.....which is good.   So, no need for an extra button.   

@jkleinsc

adeolabadmus:
Fixes #996 

- Changed `Edit` report to `View`
- Added `Print` Button to visit page


app/visits/edit/controller.js
```diff
@@ -516,6 +522,22 @@ export default AbstractEditController.extend(AddNewPatient, ChargeActions, Diagn
       this.send('update', true);
     },
 
+    printReport(report) {
```
jkleinsc: I'm not too keen on how this was implemented.  I think a better approach would be when clicking on the Print button, you are directed to the report page and print automatically fires from that page.  You can use the {{print-this}} component to automatically  print the page on render, so if you pass in a queryParam, in the template you can add the following to auto print:
```hbs
{{#if printQueryParam}} 
  {{print-this}}
{{/if}}
```

app/components/print-this.js
```diff
@@ -1,8 +1,26 @@
 import Ember from 'ember';
+
+const {
+  get,
+  isEmpty
+} = Ember;
+
 export default Ember.Component.extend({
+  print: null,
```
jkleinsc: print should default to true; otherwise invoice printing won't work unless you change that template.  I would prefer to change the default to true.

PR #1009, opened 2017-03-21
add custom form for labs

### LINKED ISSUE #65: Result form(s) for labs
Right now there is a generic result field for labs.  At some point we should support generation of result forms like the following:
screen shot 2015-04-06 at 11 41 05 am

baoqchau:
Fixes #65 

- Add custom form for lab's report.

### DIFF
--- a/app/labs/edit/route.js
+++ b/app/labs/edit/route.js
@@ -2,7 +2,6 @@ import AbstractEditRoute from 'hospitalrun/routes/abstract-edit-route';
 import AddToPatientRoute from 'hospitalrun/mixins/add-to-patient-route';
 import ChargeRoute from 'hospitalrun/mixins/charge-route';
 import Ember from 'ember';
-import moment from 'moment';
 import PatientListRoute from 'hospitalrun/mixins/patient-list-route';
 import { translationMacro as t } from 'ember-i18n';
 
@@ -11,6 +10,7 @@ export default AbstractEditRoute.extend(AddToPatientRoute, ChargeRoute, PatientL
   modelName: 'lab',
   newTitle: t('labs.newTitle'),
   pricingCategory: 'Lab',
+  customForms: Ember.inject.service(),
 
   actions: {
     returnToAllItems() {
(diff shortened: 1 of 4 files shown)


app/labs/edit/route.js
```diff
@@ -11,13 +11,22 @@ export default AbstractEditRoute.extend(AddToPatientRoute, ChargeRoute, PatientL
...
   pricingCategory: 'Lab',
+  customForms: Ember.inject.service(),
 
   actions: {
     returnToAllItems() {
       this.controller.send('returnToAllItems');
     }
   },
 
+  getCustomFormData() {
```
jkleinsc: Looks like this code isn't used, but you need this code for forms that are set to "Always Include".  Custom forms that are set to this get added automatically on new items.

app/labs/edit/template.hbs
```diff
@@ -42,6 +42,7 @@
     {{#if canComplete}}
       {{em-input property="result" label=(t 'labels.result') class="test-result-input"}}
     {{/if}}
+    {{custom-form-manager model=model formType="lab"}}
```
jkleinsc: I would move this down after notes

app/locales/en/translations.js
```diff
@@ -175,7 +175,8 @@ export default {
         textarea: 'Large Text',
         visitFormType: 'Visit',
         opdReportFormType: 'Outpatient Report',
-        dischargeReportFormType: 'Discharge Report'
+        dischargeReportFormType: 'Discharge Report',
+        labFormType: 'Lab Form Type'
```
jkleinsc: Translated text should just be 'Lab', not 'Lab Form Type'

app/labs/edit/route.js
baoqchau: So how should I fix this @jkleinsc ? You have any suggestion?

jkleinsc: Take a look at how visit does this:
https://github.com/HospitalRun/hospitalrun-frontend/blob/master/app/visits/edit/route.js#L35-L43

Basically what you need to do is to create a newData object and make sure it includes a custom forms object and then you return  customForms.setDefaultCustomForms(['lab'], newData) from the getNewData function.

jkleinsc:
Looks good to me @baoqchau.  Thanks for the PR!

PR #1017, opened 2017-03-23
Fix 1014

### LINKED ISSUE #1014: Discharge report - use the examiner field from the visit.
**Expected behavior:**

The discharge report should pull the "examiner" field off the associated admission visit.   So, on the report, it is a read only field. 

**Actual behavior:**

The report is presenting a drop-down to the user labeled "surgeon"

**Steps to reproduce:**

1. create a visit of type admission for a patient
2.  create a new discharge report, view the surgeon field

**OS and Browser:**

@jkleinsc

adeolabadmus:
Fixes #1014


app/locales/en/translations.js
```diff
@@ -1241,7 +1241,8 @@ export default {
         author: 'Author'
       },
       visitType: 'Visit Type',
-      examiner: 'Visit Examiner',
+      visitExaminer: 'Visit Examiner',
```
jkleinsc: We don't need two different labels here.  Just use Examiner.

app/visits/reports/edit/template.hbs
```diff
@@ -55,7 +55,7 @@
 
           {{#if visit.examiner }}
             <div class="ps-info-group test-examiner">
-              <p><b class="test-examiner-label">{{t 'reports.labels.examiner' }}:</b>
+              <p><b class="test-examiner-label">{{t 'reports.labels.visitExaminer' }}:</b>
```
jkleinsc: use reports.labels.examiner instead of reports.labels.visitExaminer.

app/visits/reports/edit/template.hbs
```diff
@@ -181,13 +181,12 @@
             <span class="test-visit-discharge-date-data">{{date-format model.reportDate format="DD/MM/YYYY"}}</span>
```
jkleinsc: I didn't notice this before -- this shouldn't be specifying a date format, it should use the default for the data-helper, which is the locale specific date.  Please take off the ```format="DD/MM/YYYY"```

adeolabadmus:
@jkleinsc Changes have been effected.

jkleinsc:
Looks good to me @adeolabadmus. Thanks for updating!  I'll merge it in.

PR #1020, opened 2017-03-24
Fix 1016

### LINKED ISSUE #1016: Vitals - include the user who entered the information
This is an enhancement, but I think it addresses a gap in the system currently.

Vitals (captured during a visit) should be similar to visit notes; they should capture the user (the logged -in user) who is entering the vitals information.   It will be common that vitals are take by a nurse or other clinician prior to the patient being seen by the doctor.   Therefore, the person who took the vitals should be captured.

adeolabadmus:
Fixes #1016 


app/visits/edit/template.hbs
```diff
@@ -226,6 +227,7 @@
                 </tr>
                 {{#each model.vitals as |vital|}}
                   <tr>
+                    <td>{{vital.modifiedBy}}</td>
```
jkleinsc: modifiedBy is the user's id, not the users name.

adeolabadmus: @jkleinsc, Actually, it is the user's display name.

The `modifiedBy` attribute is set here: https://github.com/HospitalRun/hospitalrun-frontend/blob/master/app/models/abstract.js#L86

adeolabadmus: @jkleinsc Oh! Ignore my comment above, I just saw your other detailed comment.

jkleinsc: @adeolabadmus hold on... you might be right here.  modifiedBy used to be the users id, but it looks like it is now user display name.  Though we probably want the user who added, not who last modified, which would require an additional attribute.

jkleinsc: I am going to merge this in with the vitals using the modifiedBy field.  I think that field is sufficient

PR #1030, opened 2017-03-26
Update initial setup steps and sample config

ninjacatt:
Fixes initial setup documents (Sorry there isn't bug tracking this yet)
While I were going over the readme.md to setup HR for the first time, I noticed some small information that missing and mismatch. So propose this update to help others if possible.

- Update readme to include install using npm and bower
- Update sample config to use the hradmin account which is being used in default script to create couchdb admin and in the readme.

### DIFF
--- a/server/config-example.js
+++ b/server/config-example.js
@@ -4,7 +4,7 @@ var config = {
   couchDbPort: '5984',
   couchDbUseSsl: false,
   couchDbChangesSince: 'now',
-  couchAdminUser: 'couchadmin',
+  couchAdminUser: 'hradmin',
   couchAdminPassword: 'test',
   googleClientId: 'FOR GOOGLE SSO; GOOGLE CLIENT ID GOES HERE',
   googleClientSecret: 'FOR GOOGLE SSO; GOOGLE CLIENT SECRET GOES HERE',
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ To install the frontend please do the following:
 ## Start
 To start the frontend please do the following:
 
+- Make sure you already ran `npm install` then `bower install` to get all dependencies.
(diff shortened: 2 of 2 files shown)


BogdanAlexandru:
Note that `npm install` and `bower install` are part of `script/bootstrap` which is executed in the installation steps.

ninjacatt:
Ah good point. I went over the server script but completely missed the bootstrap one. I could put a note to say "make sure have those installed first" or just recall this PR all together :)
		

		
	





On Sun, Mar 26, 2017 at 1:46 PM -0700, "Bogdan Alexandru Marginean" <notifications@github.com> wrote:












Note that npm install and bower install are part of script/bootstrap which is executed in the installation steps.

server/config-example.js
```diff
@@ -4,7 +4,7 @@ var config = {
   couchDbPort: '5984',
   couchDbUseSsl: false,
   couchDbChangesSince: 'now',
-  couchAdminUser: 'couchadmin',
+  couchAdminUser: 'hradmin',
```
jkleinsc: This is intentionally couchadmin, not hradmin.

There are two users that the script sets up:
1) _couchadmin_ - CouchDB admin.  This is a [CouchDB server administrator](http://docs.couchdb.org/en/2.0.0/config/auth.html?highlight=Admin%20party#server-administrators) which is a CouchDB super user.  This is used by hospitalrun-server-routes to do some user setup that a regular user can't do.
2) _hradmin_ - This is the default user to login to HospitalRun with.  This user is not a CouchDB server admin.

jkleinsc:
I am going to close this PR.   @ninjacatt if you want to update the readme to reference script/bootstrap in the start section, feel free to open a new PR with that change.

server/config-example.js
sukima: Can this be added to the README. The difference between `hradmin` and `couchadmin` really confused me (I didn't even know there was a `couchadmin`) till I saw this comment.

PR #1041, opened 2017-03-31
Address issue #390, Adjustment types for inventory need to be localized

### LINKED ISSUE #390: Adjustment types for inventory need to localized
**Expected behavior:** Inventory adjustment types display localized text.  

**Actual behavior:** Inventory adjustment types display hard coded text.  Right now values displayed for adjustment types are hard coded in app/mixins/inventory-adjustment-types.js

**Steps to reproduce:** Edit an inventory item and click on the adjust button.  The adjustment type dropdown and the button at the bottom of the modal are being pulled directly from the mixin.

One thing to keep in mind with this issue is that the 'Type' values in the mixin is used in the code and gets saved to to the database, so those values should not be localized, but anywhere we display those "type" values we need to localize them.

sweeneydavidj:
Fixes #390.

There are two issues, which are related:

1. Localization of drop-down
Added translation keys to app/mixins/inventory-adjustment-types.js. So now the drop-down list for Adjustment Types is translated. Added English translations. I didn't update the files for the other languages - I'm not sure of the process for getting translation text in multiple languages - let me know if you'd like me to do something more about that.

2. Localization of the 'submit' button.
In app/inventory/adjust/controller.js I use the the Transaction Type to find a localized Adjustment Type name.

### DIFF
--- a/app/locales/en/translations.js
+++ b/app/locales/en/translations.js
@@ -567,6 +567,13 @@ export default {
     }
   },
   inventory: {
+    adjustmentTypes: {
+      add: 'Add',
+      remove: 'Remove',
+      returnToVendor: 'Return To Vendor',
+      return: 'Return',
+      writeOff: 'Write Off'
+    },
     edit: {
       cost: 'Cost Per Unit:',
       delivered: 'Delievered To:',
--- a/app/locales/en/translations.js
+++ b/app/locales/en/translations.js
@@ -567,13 +567,6 @@ export default {
     }
   },
   inventory: {
-    adjustmentTypes: {
-      add: 'Add',
-      remove: 'Remove',
-      returnToVendor: 'Return To Vendor',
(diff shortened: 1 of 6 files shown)


app/locales/en/translations.js
```diff
@@ -553,6 +553,13 @@ export default {
     }
   },
   inventory: {
+    adjustmentTypes: {
```
jkleinsc: These values should be under inventory.labels, eg:
```js
inventory: {
  ...
  labels: {
     ...
     addAdjustmentType 'Add',
     removeAdjustmentType: 'Remove',
     ...
  }
}

sweeneydavidj:
@jkleinsc Thanks for the review. I've reorganized the localization values as suggested.

jkleinsc:
@sweeneydavidj looks good to me.  Thanks for the PR!

PR #1052, opened 2017-04-08
Fix user on example config

ssh24:
- In the docs, the default username is: `hradmin`. However, on the example-config file it defaults the username to be `couchadmin`. To keep the docs consistent with the code, I updated the username on `config-example.js`

### DIFF
--- a/server/config-example.js
+++ b/server/config-example.js
@@ -4,7 +4,7 @@ var config = {
   couchDbPort: '5984',
   couchDbUseSsl: false,
   couchDbChangesSince: 'now',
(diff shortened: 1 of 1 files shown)


jkleinsc:
@ssh24 
The user in the config is intentionally couchadmin, not hradmin.

There are two users that the couchdb script sets up:

**couchadmin** - CouchDB admin. This is a CouchDB server administrator which is a CouchDB super user. This is used by hospitalrun-server-routes to do some user setup that a regular user can't do.
**hradmin** - This is the default user to login to HospitalRun with. This user is not a CouchDB server admin.

PR #1056, opened 2017-04-17
add review and print button for cashier screen

### LINKED ISSUE #303: Cashier screen

baoqchau:
Fixes #303 .

- add "Review" button for cashier to review invoices
- add "Print" button for cashier to print invoices
screenshot_20170416_181910

However, I still need to refactor the code because the code for `cashier` route is kind of the same like `index` route except for the button.
Updated: I have refactored the code for `print invoice`.


baoqchau:
@jkleinsc What I would like to do is that we will change the `Invoice` subnav into `Cashier Screen` if the user roles is `cashier` and `Invoice` if the role is different. Also, the buttons will also be added based on the user roles. In this way, it will look cleaner for the cashier.

jkleinsc:
@baoqchau that sounds good to me.  I would change the label to `Cashier` instead of `Cashier Screen`.  You also can extend the invoice controller/route for use in the cashier controller/route if the code is pretty similar, eg:
```js
import InvoiceRoute from 'hospitalrun/invoices/route';
export default InvoiceRoute.extend({
...
});
```

baoqchau:
After modifying, this is the `invoices/index` for `admin` roles. 
screenshot_20170422_161853
And this is the screen for the `cashier' roles.
screenshot_20170422_161825
I think that we don't need to change the title to "Cashier" because when we click each type of `Invoices`, the title will change with it. In addition, I have 2 questions about the requirements for the cashier screen:
1/ Should the cashier allow to edit invoices? If not, we should make the bill `unclickable`.
2/ Should the cashier allow to add new invoices? If not, we should hide the button `+ new invoice`

tangollama:
Thanks @baoqchau. We'll review this week and (hopefully) get this merged into master.

baoqchau:
@tangollama @jkleinsc Is there anything I can do to help with this travis CI build failure? Looks like the error came from the container using to build the CI.

baoqchau:
@jkleinsc done. Is there anything else you want me to change?

tangollama:
@baoqchau I'm going to update this with the latest and (assuming all the tests pass, which they should), I'll merge this today. Thanks for your patience.

baoqchau:
@tangollama any updates on this yet? Just want to check up.

jkleinsc:
Looks good to me @baoqchau.  We will merge it in.

PR #1064, opened 2017-04-23
Modified visit model

hsorellana:
This is a correction of the visit model. I had to open a new PR since i deleted the fork and all changes locally made are gone.

I suggest the PR #919 should be deleted since its commit are in this one, and i can keep tracking the work here.
cc @HospitalRun/core-maintainers

PR #1065, opened 2017-04-23
Modified medication model

hsorellana:
Modifications requested, in a new PR, cause of what i posted on #1064 

Same here, suggest that #913 PR should be deleted.


app/models/medication.js
```diff
@@ -109,27 +102,28 @@ export default AbstractModel.extend(CanEditRequested, DateFormat, MedicationDeta
...
       },
       acceptance: {
         accept: true,
         if(object) {
-          let isFulfilling = object.get('isFulfilling');
-          let requestQuantity = parseInt(object.get('quantity'));
+          let isFulfilling = get(object, 'isFulfilling');
+          let requestQuantity = parseInt(get(object, 'quantity'));
           let quantityToCompare = null;
+
```
jkleinsc: There are spaces here that are causing ESLint to fail.  Can you remove them?

jkleinsc:
Thanks for the update on the PR @hsorellana!

PR #1087, opened 2017-05-19
Normalize line endings within readme

abactel:
Previously, lines were ended in a variety of ways. Most significantly,
the use of periods and blanks was mixed. I've added periods and colons
to appropriate locations.

-    Change all line endings to full stops or colons (mainly addresses lists).


jglovier:
@abactel thanks for this! It's always helpful to make docs clearly. :metal:

Issue #1094, opened 2017-05-24
A button for switch between language...


jujes:
**Expected behavior:**
option to switch from English to Spanish
**Actual behavior:**
No Spanish language as choice

Hi, someone know how I can make available the Spanish has a choice for user interface?

thanks in advance,

tangollama:
Thanks @jujes. This is a missing feature and one that someone in the community could (relatively) quickly tackle. I'll label appropriately.

wildaces215:
Iss this open I would like to look into it!

wildaces215:
I look into this!

richiethomas:
Should the language settings be stored as an attribute of the user, so they persist with each user session?

tangollama:
A global config for the default and a user preference are in order. 

Typed or dictated imperfectly on my iPhone (717.385.9970)

tchan:
@richiethomas 
Hopefully these links will be helpful, although I am unsure
https://eviltrout.com/2013/11/24/i18n-in-ember.html
https://github.com/jamesarosen/ember-i18n

richiethomas:
@tchan yeah looks like that's currently a dependency of the project.  I'll definitely need to read through the package documentation as part of playing the story.

@tangollama since there's no design spec to work off, is there a certain location where the language toggle should go?  Or should I just come up with something and submit it for consideration?  My first thought was adding a select element to the dropdown that appears when clicking the "Settings" gear.

tangollama:
I think that's a great recommendation on the gear location. Perhaps @jglovier can offer some guidance as well.

richiethomas:
Thanks @tangollama.  Follow-up question- in the past I've used an npm package called `ember-power-select` for use cases like this.  It's powerful and well-maintained.  I'm thinking about using it in the implementation of the language select dropdown, but since it would be a new dependency I wanted to run it by y'all first.

More info here- https://github.com/cibernox/ember-power-select

jkleinsc:
@richiethomas I would prefer to not add another select component to the project.  We currently use ember-rapid-forms which has a select component (em-select) and there also is a another component, ember-select-list, in the project.

richiethomas:
@jkleinsc gotcha, wasn't aware of those other 2 dependencies.  I'll investigate those instead.

richiethomas:
[...]

PR #1109, opened 2017-06-22
Lang select final

richiethomas:
Addresses #1094 

- Add a dropdown-select element to the settings modal, viewable by clicking the gear icon in the navbar
- The dropdown includes a hard-coded list of languages (for now just 2- English and French), and upon making a selection the language of any text under Handlebars i18n control will immediately update to the new language

NOTE- there are still many elements in the Handlebars views that are either:

a) hard-coded as English, or
b) controlled by i18n but require a page refresh because the i18n logic is in the component, not the view template

In the first case, these will need to be refactored to be controlled by the i18n package.  That's pretty straightforward.

In the 2nd case, this is a little more difficult.  The reasons for putting the i18n logic vary from component to component, so these refactorings will need to be addressed on a case-by-case basis.  For an example, see the logic in `app/inventory/route.js`, which includes a translation for the `inventory received` button text.  There is a function `additionalButtons` which contains information on the button action, button text, classes, etc. which seems to be should be extracted into the view, but this case and others like it require a bit of surgery and I thought it best to get this PR up first.

NB: I was able to extract the button text in the above example from hard-coded english to a translation, but when I try to make it a computed property, the button text renders as empty in the browser.  Not sure why that is, but hopefully the issue will be irrelevant if and when I can extract the i18n logic into the view itself using the "t" view helper.

I'd appreciate any feedback on additional test coverage desired, or on refactoring the code itself to make it more idiomatic for Ember, but any feedback will be considered. :-)

Also, I'll add screenshots of the functionality as it stands now to the Github issue and the comments below, but it may also be advisable to pull down the code from the branch and test it manually on your local machine, before giving a final go-ahead.

### DIFF
--- a/app/admin/lookup/route.js
+++ b/app/admin/lookup/route.js
@@ -5,7 +5,7 @@ const { computed } = Ember;
 
 export default AbstractIndexRoute.extend({
   hideNewButton: true,
-  pageTitle: computed('i18n', () => {
+  pageTitle: computed('i18n.locale', () => {
     return t('admin.lookup.pageTitle');
   }),
   model() {
--- a/app/admin/lookup/route.js
+++ b/app/admin/lookup/route.js
@@ -1,8 +1,13 @@
 import AbstractIndexRoute from 'hospitalrun/routes/abstract-index-route';
 import { translationMacro as t } from 'ember-i18n';
+import Ember from 'ember';
+const { computed } = Ember;
+
 export default AbstractIndexRoute.extend({
   hideNewButton: true,
-  pageTitle: t('admin.lookup.pageTitle'),
+  pageTitle: computed('i18n', () => {
+    return t('admin.lookup.pageTitle');
+  }),
   model() {
     return this.store.findAll('lookup').catch((error) => this.send('error', error));
   },
--- a/app/admin/textreplace/route.js
+++ b/app/admin/textreplace/route.js
@@ -4,7 +4,7 @@ import Ember from 'ember';
 const { computed } = Ember;
 
 export default AbstractIndexRoute.extend({
-  pageTitle: computed('i18n', () => {
+  pageTitle: computed('i18n.locale', () => {
     return t('admin.textReplacements.pageTitle');
   }),
   hideNewButton: true,
--- a/app/admin/textreplace/route.js
+++ b/app/admin/textreplace/route.js
@@ -1,8 +1,12 @@
 import AbstractIndexRoute from 'hospitalrun/routes/abstract-index-route';
 import { translationMacro as t } from 'ember-i18n';
+import Ember from 'ember';
+const { computed } = Ember;
 
 export default AbstractIndexRoute.extend({
-  pageTitle: t('admin.textReplacements.pageTitle'),
+  pageTitle: computed('i18n', () => {
+    return t('admin.textReplacements.pageTitle');
+  }),
   hideNewButton: true,
 
   model() {
--- a/app/appointments/search/route.js
+++ b/app/appointments/search/route.js
@@ -10,7 +10,7 @@ export default AppointmentIndexRoute.extend(DateFormat, {
   editReturn: 'appointments.search',
   filterParams: ['appointmentType', 'provider', 'status'],
   modelName: 'appointment',
-  pageTitle: computed('i18n', () => {
+  pageTitle: computed('i18n.locale', () => {
     return t('appointments.searchTitle');
   }),
 
--- a/app/appointments/search/route.js
+++ b/app/appointments/search/route.js
@@ -4,11 +4,15 @@ import Ember from 'ember';
 import moment from 'moment';
 import { translationMacro as t } from 'ember-i18n';
 
+const { computed } = Ember;
+
 export default AppointmentIndexRoute.extend(DateFormat, {
   editReturn: 'appointments.search',
   filterParams: ['appointmentType', 'provider', 'status'],
   modelName: 'appointment',
-  pageTitle: t('appointments.searchTitle'),
+  pageTitle: computed('i18n', () => {
+    return t('appointments.searchTitle');
+  }),
 
   queryParams: {
     appointmentType: { refreshModel: true },
--- a/app/appointments/theater/route.js
+++ b/app/appointments/theater/route.js
@@ -1,10 +1,17 @@
 import AppointmenCalendarRoute from 'hospitalrun/appointments/calendar/route';
 import { translationMacro as t } from 'ember-i18n';
+import Ember from 'ember';
+
+const { computed } = Ember;
 
 export default AppointmenCalendarRoute.extend({
   editReturn: 'appointments.theater',
-  newButtonText: t('appointments.buttons.scheduleSurgery'),
-  pageTitle: t('appointments.titles.theaterSchedule'),
+  newButtonText: computed('i18n', () => {
+    return t('appointments.buttons.scheduleSurgery');
+  }),
+  pageTitle: computed('i18n', () => {
+    return t('appointments.titles.theaterSchedule');
+  }),
 
   _modelQueryParams(params) {
     let queryParams = this._super(params);
--- a/app/appointments/theater/route.js
+++ b/app/appointments/theater/route.js
@@ -6,7 +6,7 @@ const { computed } = Ember;
 
 export default AppointmenCalendarRoute.extend({
   editReturn: 'appointments.theater',
-  newButtonText: computed('i18n', () => {
+  newButtonText: computed('i18n.locale', () => {
     return t('appointments.buttons.scheduleSurgery');
   }),
   pageTitle: computed('i18n', () => {
--- a/app/appointments/theater/route.js
+++ b/app/appointments/theater/route.js
@@ -9,7 +9,7 @@ export default AppointmenCalendarRoute.extend({
   newButtonText: computed('i18n.locale', () => {
     return t('appointments.buttons.scheduleSurgery');
   }),
-  pageTitle: computed('i18n', () => {
+  pageTitle: computed('i18n.locale', () => {
     return t('appointments.titles.theaterSchedule');
   }),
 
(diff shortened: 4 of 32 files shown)


richiethomas:
before clicking the gear
clicking the gear and seeing the dropdown
selecting a language
language changed from english to french note the change in sidebar and table header translations but page title has not changed
after page refresh page title has changed too

richiethomas:
P.S.- ignore the `<script>alert("Hi")</script>` in the Patient Name column.  I was doing some security testing. :-p

app/imaging/index/route.js
```diff
@@ -1,8 +1,14 @@
 import { translationMacro as t } from 'ember-i18n';
 import AbstractIndexRoute from 'hospitalrun/routes/abstract-index-route';
+import Ember from 'ember';
+
+const { computed } = Ember;
+
 export default AbstractIndexRoute.extend({
   modelName: 'imaging',
-  pageTitle: t('imaging.pageTitle'),
+  pageTitle: computed('i18n', () => {
```
jkleinsc: This should be `computed('i18n.locale'....`

app/inventory/route.js
```diff
@@ -1,14 +1,16 @@
+import { translationMacro as t } from 'ember-i18n';
 import AbstractModuleRoute from 'hospitalrun/routes/abstract-module-route';
 import FulfillRequest from 'hospitalrun/mixins/fulfill-request';
 import InventoryId from 'hospitalrun/mixins/inventory-id';
 import InventoryLocations from 'hospitalrun/mixins/inventory-locations'; // inventory-locations mixin is needed for fulfill-request mixin!
 export default AbstractModuleRoute.extend(FulfillRequest, InventoryId, InventoryLocations, {
   addCapability: 'add_inventory_item',
+  buttonText: t('navigation.subnav.inventoryReceived'),
```
jkleinsc: This buttonText should use a separate localization entry, not reuse the one from the subnav (the text is intentionally lowercase for the button).

app/routes/application.js
```diff
@@ -106,6 +106,10 @@ let ApplicationRoute = Route.extend(ApplicationRouteMixin, ModalHelper, SetupUse
   afterModel() {
     set(this.controllerFor('navigation'), 'allowSearch', false);
     $('#apploading').remove();
+    this.get('config.configDB').get('current_user').then((user) => {
+      let language = user.i18n || 'en';
```
jkleinsc: This value isn't set anywhere.

app/admin/lookup/route.js
```diff
@@ -1,8 +1,13 @@
 import AbstractIndexRoute from 'hospitalrun/routes/abstract-index-route';
 import { translationMacro as t } from 'ember-i18n';
+import Ember from 'ember';
+const { computed } = Ember;
+
 export default AbstractIndexRoute.extend({
   hideNewButton: true,
-  pageTitle: t('admin.lookup.pageTitle'),
+  pageTitle: computed('i18n', () => {
```
richiethomas: Fixed.

app/admin/textreplace/route.js
```diff
@@ -1,8 +1,12 @@
 import AbstractIndexRoute from 'hospitalrun/routes/abstract-index-route';
 import { translationMacro as t } from 'ember-i18n';
+import Ember from 'ember';
+const { computed } = Ember;
 
 export default AbstractIndexRoute.extend({
-  pageTitle: t('admin.textReplacements.pageTitle'),
+  pageTitle: computed('i18n', () => {
```
richiethomas: Fixed.

app/appointments/search/route.js
```diff
@@ -4,11 +4,15 @@ import Ember from 'ember';
...
 import { translationMacro as t } from 'ember-i18n';
 
+const { computed } = Ember;
+
 export default AppointmentIndexRoute.extend(DateFormat, {
   editReturn: 'appointments.search',
   filterParams: ['appointmentType', 'provider', 'status'],
   modelName: 'appointment',
-  pageTitle: t('appointments.searchTitle'),
+  pageTitle: computed('i18n', () => {
```
richiethomas: Fixed

app/appointments/theater/route.js
```diff
@@ -1,10 +1,17 @@
...
 import { translationMacro as t } from 'ember-i18n';
+import Ember from 'ember';
+
+const { computed } = Ember;
 
 export default AppointmenCalendarRoute.extend({
   editReturn: 'appointments.theater',
-  newButtonText: t('appointments.buttons.scheduleSurgery'),
-  pageTitle: t('appointments.titles.theaterSchedule'),
+  newButtonText: computed('i18n', () => {
```
richiethomas: Fixed

app/appointments/today/route.js
```diff
@@ -1,10 +1,16 @@
...
 import { translationMacro as t } from 'ember-i18n';
+import Ember from 'ember';
+
+const { computed } = Ember;
+
 export default AppointmentIndexRoute.extend({
   editReturn: 'appointments.today',
   modelName: 'appointment',
-  pageTitle: t('appointments.todayTitle'),
+  pageTitle: computed('i18n', () => {
```
richiethomas: Fixed

app/imaging/index/route.js
richiethomas: Fixed

app/components/language-select.js
```diff
@@ -0,0 +1,39 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  config: Ember.inject.service(),
+  i18n: Ember.inject.service(),
+  languageOptions: Ember.computed(function() {
+    return Object.keys(this.get('languageMap'));
+  }),
+  onFinish: null,
+  languageMap: Ember.computed(() => {
```
richiethomas: Fixed.  Duplicated code removed and language names localized.  One callout- I was able to remove the `value` attribute from the select-list without any apparent adverse effects.  Not sure why this was the case, since I thought `value` maps to the `value` attribute of the `option` tag.  Does this get overridden if the `optionValuePath` and `optionLabelPath` attrs are used instead?

app/components/language-select.js
```diff
@@ -0,0 +1,39 @@
...
+    return {
+      'English': 'en',
+      'French': 'fr'
+    };
+  }),
+  selectedLanguage: null,
+
+  _setUserLanguageChoice(language) {
```
richiethomas: Hmmm, I thought the record being written was in fact being used, specifically by the `afterModel` hook in `app/routes/application.js`.  My intent here in `_setUserLanguageChoice` was to write the user's language preference to the database, and then if the user logs out and later logs back in, their language choice is retrieved from the DB in the above-mentioned `afterModel` hook.  I tested it in the browser and the app's behavior is what I expected, but am I confused about how the behavior works under the hood?

app/routes/application.js
richiethomas: See my above comment from `app/components/language-select.js`.

app/components/language-select.js
richiethomas: Also, I added localized English/Spanish/French values for the labels "English", "Spanish", and "French", but stopped short of adding the respective keys for all supported languages (i.e. the Russian translation for the label "Urdu", etc.).

app/inventory/route.js
richiethomas: Removed this as it was mostly a nice-to-have and I don't want to block the story.

app/routes/application.js
jkleinsc: `_setUserLanguageChoice(language)` saves a new record in the configDB, with the id of the record being the user, but here you are just fetching the current_user, which has an id of 'current_user', not the id of the user.  Does that make sense?

richiethomas: You mentioned `_setUserLanguageChoice(language)` saves a new record in the configDB.  I had a different understanding of how it works.  I believe `_setUserLanguageChoice(language)` first fetches the current_user from the configDB, i.e. `configDB.get('current_user')`.  That returns a promise, and when that's resolved successfully it passes a `user` object to the success callback.  That callback takes the user object and uses its `_id` attribute to send a `put` call to the `configDB`.  That `put` call looks like this:

    configDB.put({
            i18n: language,
            _id: user._id,
            _rev: user._rev,
            value: user.value
          });

Because I specified the `_id` attribute of the user, my understanding is that PouchDB will overwrite the existing record's attributes (i.e. an update operation) instead of creating a new record.  My intent was to follow the guide here:

https://pouchdb.com/guides/documents.html#updating-documents–correctly

As a tangent to this conversation, to make my intent clearer I refactored the code a bit in `_setUserLanguageChoice(langauge)` (now called `_setUserLanguage(language)` for succinctness).

I'm still learning how both HospitalRun and PouchDB work, so I could be wrong.  Happy to hop on a Slack chat or something if it's easier to talk that way.

jkleinsc: When you call `configDb.get('current_user')` it pulls a record with an `_id === 'current_user'`, but when you save the record with `user._id` the id of that record is going to be something like `org.couchdb.user:couchadmin`, not `current_user`.  Does that make sense?

app/components/language-select.js
```diff
@@ -19,12 +19,8 @@ export default Ember.Component.extend({
   _setUserLanguage(language) {
     let configDB = this.get('config.configDB');
     configDB.get('current_user').then((user) => {
-      configDB.put({
-        i18n: language,
-        _id: user._id,
-        _rev: user._rev,
-        value: user.value
-      });
+      user.i18n = language;
```
jkleinsc: This is the change I was expecting.  This will work!

app/components/language-select.js
jkleinsc: @richiethomas `value` is used to set the currently selected value of the drop down.  It is a way of setting which `option` is `selected`.

jkleinsc:
@richiethomas I'm going to go ahead and merge this in.  Thanks for your work on this!

PR #1120, opened 2017-07-05
change label and add completedBy field in imaging page

### LINKED ISSUE #1033: Imaging requests - changes to notes, field order and completed by
This is an enhancement request

1.  Add a "request notes" field, right after the image type selection field.   This would be where the ordering doctor specifies the details of the request (i.e.   please take an image of the upper left arm)

2. Rename the existing notes field to be "results notes"   This will be used by either the tech taking the image, or the radiologist reading the image, to add notes, observations, etc.   

3. A system populated field named "completed by" should be added when the image request is completed.  

It is important to note that technicians take images, radiologist interpret them.   So, who completes the image should be captures in all cases, but the radiologist that interprets and comments on the image is a second optional piece of information

the relevant field order, top to bottom should be:

1. image type
2. request notes
3. radiologist
4. results notes
5. completed by:

baoqchau:
Fixes #1033 .

- Change labels of `Result` into `Result Notes` and `Notes` into `Requested Notes`
- Add new field `completedBy` in `imaging` models as well as on the `Edit` page of `imaging`
screenshot_20170704_173311


baoqchau:
I did not make the `completedBy` fields in `Edit` page of imaging a populated field, because I don't know what kind of user will complete it (imaging technician, doctor, or radiologist). However, if we can make it a populated field as requested in issue #1033 , what kind of user should I populated it with?

jkleinsc:
@baoqchau completed by should be set to the user who clicks on the complete button on the request.  If you look at how requestedBy is populated in the code you should be able to do the same for completedBy (eg completedBy is current user when complete button is clicked).

baoqchau:
I add `Completed By` column in `completed` page of `imaging`, which will be easier to track down who completed that request.

baoqchau:
I added `Completed By` column in `Completed` page of `Imaging`, which make it easier to track down who completed that imaging request.
screenshot_20170706_110929

PR #1121, opened 2017-07-10
Added photos to visits and procedures (#1106)

### LINKED ISSUE #1106: Add photos specifically to a procedure and visit
**Expected behavior:**
Currently, the system supports attaching multiple photos to a patient, but there's no granularity down to the Visit or Procedure level. Attaching photos to a visit and procedure and having them specifically surface in those interfaces would be preferable from a usability standpoint.

**Actual behavior:**
Currently, photos are attached only to patients and not (optionally) categorized by visit or procedure.

- [ ] Add optional meta data to photos that allow visit and procedure ID's to be added by the system
- [ ] Add photo uploads to the Visit screen
- [ ] Add photo display to the Visit screen
- [ ] Add photo uploads to the Procedure screen
- [ ] Add photo display to the Procedure screen

**Steps to reproduce:**
N/A

N/A

**OS and Browser:**
N/A

ArtificialQualia:
Fixes #1106 

Note: I added the functionality as I mentioned in #1106, which is to say when you upload a photo through the visits or procedure screen, the photo will then show up in the screen you uploaded it from AS WELL AS the patient photo section.  This allows access to all of a patient's photos, regardless of where they were uploaded from, to be accessible in the same place.  Of course, only the photos that were attached to a specific visit or procedure will show in their own respective views.

- Updated photo model with visit and procedure field
- Changed photo controller to have a editController set by whomever is updating the photos, rather than a hardcoded 'patients/edit'
- Added pouch views for photos_by_visit and photos_by_procedure
- Added photo functionality to route and controllers for visits and procedures
- Added photo tab to visit template
- Added photo view to procedure template
- Updated test fixtures with new pouch views


ArtificialQualia:
I think the automated tests failed due to how I rebuilt the test fixtures.  I regenerated them to remove the warnings about the new pouch_views I added, but it seems I did something wrong (tried running the commands as described in the contributing doc).

The feature works, but I'll try to explore why regenerating the fixtures files is causing the build to fail.

ArtificialQualia:
The tests do pass when I use the original test/fixtures files by the way, just with a lot of warnings about my new pouch_views

ArtificialQualia:
I see now that the original fixtures contained test data that was not present in my dumps.

I have manually added the two new views to all the fixtures files, and now the test are passing without warnings about my new views.

app/patients/photo/controller.js
```diff
@@ -15,7 +15,7 @@ export default AbstractEditController.extend({
   showUpdateButton: true,
 
   database: inject.service(),
-  editController: inject.controller('patients/edit'),
+  editController: null,
```
jkleinsc: This line should be removed as editController is not used from this context anymore.

ArtificialQualia:
@jkleinsc I made the changes you requested to `photo/controller.js` and reverted the changes I made to the testing fixtures.

However, regarding the tests you requested I add, those are not in my commit.  There doesn't appear to be a good way to test photo uploading programmatically.  You can't set the value of the file input directly, as that is blocked for security reasons.  I tried to do it programmatically via a manually created blob and change event, similar to [what I found here](https://stackoverflow.com/questions/29120799/acceptance-test-for-file-uploading-in-ember-cli), but that does not appear to be working in this scenario, as the file input is ignoring the change event when called programatically.

It appears that photo uploading is also not currently being tested for in `tests/acceptance/patients-test.js`, probably for similar reasons why I had issues adding it.

For your reference, here is an example of the code I tried to use for the automated testing:

```
  andThen(function() {
    click('button:contains(New Photo)');
    waitToAppear('.modal-dialog');
  });
  andThen(function() {
    assert.equal(find('.modal-title').text(), 'Add Photo', 'Add Photo dialog displays');
    $('.photo-source').prop('selectedIndex', 1).trigger('change');
    waitToAppear('.image-upload'); // Note: this is a class I added to take-photo.hbs in order for this to work
  });
  andThen(function() {
    // b64Data is a red dot.  We could make it a single pixel if we want this string to be smaller.
    let b64Data = 'iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg==';
    let byteCharacters = atob(b64Data);
    let byteNumbers = new Array(byteCharacters.length);
    for (let i = 0; i < byteCharacters.length; i++) {
      byteNumbers[i] = byteCharacters.charCodeAt(i);
    }
    let blob = new Blob([new Uint8Array(byteNumbers)], { type: 'image/png' }); // Note: for some reason I this line was giving me errors from eslint due to it thinking Uint8Array isn't a global.  This was strange to me since it is used many other places in the code without error.
    blob.name = 'image.png';
    let inputElement = $('input[type=file]');
    inputElement.triggerHandler({
      type: 'change',
      target: {
        files: {
          0: blob,
          length: 1
        }
      }
    });
  });
  andThen(function() {
    click('.modal-footer button:contains(Add)');
    waitToDisappear('.modal-dialog');
  });
```

I tried many variations of that code, but as I said, this code does not get the file input to correctly handle the change event.  This can also be tested in your javascript console through a browser, it simply does not work outside the file chooser.

tangollama:
Thx @ ArtificialQualia for the reply. @jkleinsc can you review again?

PR #1132, opened 2017-07-23
Change i18n preference selection so that it's user-specific, and persists after logout/login

### LINKED ISSUE #1094: A button for switch between language...

richiethomas:
Fixes #1094.

 - Add PouchDB preferences DB, where things like user-specific i18n preferences can be stored
 - upon language selection, user i18n preference is stored in preferences DB
 - upon login, i18n.locale is set to user-specific preference
 - when a different user logs in, their own preferred i18n setting will be used

NOTE: 
If a user selects a language other than English, then logs out, they will see a split-second flash of English when logging back in, before their previously-chosen language translations appear.  For some reason, this can be prevented if the user logs out and does a hard browser refresh from the login page, before logging back in.  If the hard refresh is performed, the user will see their i18n settings show up on the login screen for a split-second before being taken to their home screen.  I assume this is preferable, but am unable to fix the bug I initially described.

Currently there is very little test coverage for this feature.  I have commented-out partial acceptance tests but I observe consistent error message in the JS console related to missing translations.  I may not be loading the ember-i18n service correctly, although I've tried to mock it out and was unsuccessful with that strategy too.

Mostly submitting this PR as a status check, for feedback purposes.


app/components/language-select.js
```diff
@@ -3,30 +3,40 @@ import Ember from 'ember';
...
 
-  _setUserLanguage(language) {
+  _storeUserI18n(i18n) {
     let configDB = this.get('config.configDB');
+    let username;
     configDB.get('current_user').then((user) => {
-      user.i18n = language;
-      configDB.put(user);
+      username = (typeof user.value === 'string') ? user.value : user.value.name;
+      configDB.get('preferences').then((db) => {
```
jkleinsc: this line should be:
```js
configDB.get('preferences').then((preferences) => {
``` 
As it is getting a reference to a record, not the whole db.

app/services/language-preference.js
```diff
@@ -0,0 +1,35 @@
...
+    let userName;
+    configDB.get('current_user').then((doc) => {        // Set username in current_user DB
+      userName = doc.value;
+    }).then(() => {                                   // Fetch preferences DB
+      return configDB.get('preferences');
+    }).then((doc) => {                                //  Set current user's i18n preference
+      let name = (typeof userName === 'object') ? userName.name : userName;
+      let i18nPreference = (doc[name]) ? doc[name].i18n : doc.default.i18n;
+      this.set('i18n.locale', i18nPreference);
+      return Ember.RSVP.Promise.resolve();
```
jkleinsc: This could just be a return; it doesn't need to return a resolve.

app/components/language-select.js
richiethomas: Fixed.

app/services/language-preference.js
richiethomas: Fixed.

tests/acceptance/language-dropdown-test.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+});
+
+test('setting a language preference persists after logout', (assert) => {
+  runWithPouchDump('default', () => {
+    authenticateUser({ name: 'foobar', prefix: 'p2' });
+    visit('/');
+
+    andThen(() => {
+      debugger;
```
jkleinsc: Can you remove these debugger statements?

app/services/language-preference.js
```diff
@@ -0,0 +1,78 @@
...
+      preferences[username].i18n = selection;
+      configDB.put(preferences);
+      this.set('i18n.locale', selection);
+    }).catch((err) => {
+      console.log(err);
+      this._initPreferencesDB(username, selection);
+    });
+  },
+
+  // setUserI18n(selection) {
```
jkleinsc: Can you remove this commented out code?

tests/acceptance/language-dropdown-test.js
```diff
@@ -0,0 +1,52 @@
...
+      assert.equal(currentURL(), '/');
+      click('a.settings-trigger');
+    });
+    andThen(() => {
+      debugger;
+      select('.language-dropdown', 'German');
+    });
+    // andThen(() => {
```
jkleinsc: This commented out code should be removed

richiethomas:
Hey @jkleinsc, I actually don't think this is merge-able quite yet.  My local database got into a funky state and I can't guarantee it's not related to the changes in this PR (in fact I suspect it might be).  I am scheduled to pair with someone tonight to work on the issue.  Can I update you later this week when I know more?

richiethomas:
@jkleinsc OK so after extensive manual testing, I think this is ready for a merge, with some caveats:

1) The acceptance tests are a WIP, and have been commented-out and marked with "WIP" at the top of the file.  I am submitting my work now in order to unblock this feature, and will submit the acceptance tests after additional work if that's OK.

2) Also, there is code duplication in the `language-preference` service, specifically in the `setUserI18nPreference` and `storeUserI18n` methods.  This is intentional and temporary- there is already a lot of similarity in what these two methods do, with minor exceptions, so I am preparing to extract the common code that is shared between these two methods.  Let me know if you'd like me to explain further.

richiethomas:
I've reached an impasse while writing the acceptance tests for this story.  The feature involves fetching the user's stored i18n preference and setting it as the current i18n locale whenever the user logs in, and there seem to be intricacies involved in the login process in the test environment that are preventing me from writing tests as expected.

For example, I found that using the `authenticateUser` helper function doesn't actually cause the code in the `routes/application.js` file to be called.  It appears to be a shortcut for logging in a user without needing to go through the process of using the `fillIn` helper method to manually fill in the login form.

Also, I tried doing exactly the above steps (using the `fillIn` helper method), but ran into problems stubbing certain database calls.  I was able to stub the authentication response, but there appear to be DB calls after that which I'm unable to trace, and therefore unable to stub.

At this point I have to put my hand up and ask for help.  I've expended time and money hiring Ember help from Codementor.io to pair with me, and while I did manage to make incremental process, I am unable to afford the continuous expense of doing so.  I am more than happy to pair with someone more familiar with the codebase in order to share what I know about the component and also learn how to solve the problem.  But I've reached the conclusion that I've made as much progress as I can on my own.  Any help would be appreciated.

PR #1133, opened 2017-07-23
Added missing languages translations in the dropdown

tchan:
Complements/Addresses #1094.

- Added language translations for all available languages

German
Spanish and so on. 

I am a bit confused how there are two dropdowns for spanish and spanish (colombian). As far as I can tell the languages are pretty much the same and differ in dialect/slang. Would love some clarification. 
https://www.quora.com/What-are-the-major-differences-between-Colombian-Spanish-and-Spanish-from-Spain
https://www.lonelyplanet.com/thorntree/forums/americas-mexico/mexico/how-different-is-mexican-spanish-to-colombian-spanish

Also after reviewing my commit files, should I not have included my package-lock.json file (17k lines of code...)? I couldn't find anything about it in the contributing guidelines. Edit: I have now removed it from the commit files.

### DIFF
--- a/app/locales/de/translations.js
+++ b/app/locales/de/translations.js
@@ -875,15 +875,15 @@ export default {
     sectionTitle: ''
   },
   languages: {
-    en: '',
-    fr: '',
-    es: '',
-    de: '',
-    ru: '',
-    'es-co': '',
-    'pt-br': '',
-    tr: '',
-    ur: ''
+    en: 'Englisch',
+    fr: 'Französisch',
+    es: 'Spanisch',
+    de: 'Deutsche',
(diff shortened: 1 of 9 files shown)


richiethomas:
@tchan I believe you're right, and that the differences are comparable to British vs American English.  Some apps have i18n options which allow for both, a design choice which depends on the app's audience.

tchan:
Makes sense imo to get rid of it but what do you guys think?

egibar:
The Spanish and Colombian Spanish are basically the same, there are words that change or expressions but I don't think that is enough to put it as a different language. It the same for example to Argentinian Spanish.

PR #1153, opened 2017-08-30
Fixes: Navigation arrows obscure database column headings #1068

### LINKED ISSUE #1068: Navigation arrows obscure database column headings
The arrows that are used to scroll between screens of the databases in HospitalRun are placed such that sometimes they cover the headings of columns in the databases. The placement should be revised so that, in all user views, this does not occur.

For example, logged in as "hr.doctor@hospitalrun.io", navigate to the "Medication" tab. I have attached a screenshot where you can see the arrows obscuring the "Status" heading (top right of the screen).

This is using Chrome on Debian Jessie, I haven't yet tested on other OSs or browsers but I will do so and update this report accordingly.

hospitalrunarrows

EvanSays:
Fixes #1068 .

- Add class="actions" to action column in app/medication/index/template.hbs.
- Create _medication.scss and gave actions a fixed width.
- Update app.scss to reflect new _medications.scss file.

Description: We set a fixed width on the actions column to bump it over to the left, allowing the nav-paging buttons to not be hidden.

Tested: Latest version of Chrome, Safari, Firefox.

[Before]
Before

[After]
After

Not added in this solution, but something we came up with along the way, we moved the nav-paging buttons above and below the table. This would allow the tables to never be obstructed by the buttons and gives the ability for the user to change pages when at the bottom scroll. Just something to consider for future.

example 1

example 2

Contributors @jsullivan5 @the-oem @evansays

### DIFF
--- a/app/medication/index/template.hbs
+++ b/app/medication/index/template.hbs
@@ -21,7 +21,7 @@
           <td>{{medicationRequest.quantity}}</td>
           <td>{{medicationRequest.status}}</td>
           {{#if showActions}}
-            <td>
+            <td class="actions">
               <button class="btn btn-primary" {{action 'editItem' medicationRequest bubbles=false }}>{{t 'labels.fulfill'}}</button>
             </td>
           {{/if}}
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -29,6 +29,7 @@
 @import 'components/imaging';
 @import 'components/labs';
 @import 'components/loading_notice';
+@import 'components/medication';
 @import 'components/pagination';
 @import 'components/panel';
 @import 'components/patient_history';
--- a/app/styles/components/_medication.scss
+++ b/app/styles/components/_medication.scss
@@ -1,0 +1,1 @@
+.actions { width: 160px; }


tchan:
Great job @EvanSays @jsullivan5 @the-oem !

I'm personally in the camp of having the pagination buttons on the top and bottom of the table. There are more userability advantages in having it at the top and bottom and the only disadvantage I can think of is that it takes up more space on the screen which IMO as well, isn't a big deal.

Either way, awesome work!

tangollama:
Thanks @EvanSays for the PR. I've asked @jglovier to take a quick look at it today and either approve or comment.

EvanSays:
@jglovier. Thank you for the feedback. We are really happy you all liked this other method as well. We will have a pullup request up shortly with the new fix to add the buttons atop and below the table. @jsullivan5 @the-oem

EvanSays:
@jglovier Added the nav-paging buttons below and atop the table was a site-wide change we tested before. Would you prefer the fix to be scoped to the medications table only? or site-wide?

NuttasitBoonwat:
M

EvanSays:
@tangollama We're currently in the mix of finals at school, which is why we haven't been active recently. Wanted to let you know that we're planning to fix this when our workload dies down in the next week or so.

jglovier:
> We will have a pullup request up shortly with the new fix to add the buttons atop and below the table.

Feel free to just push those changes to this PR. No need to start a new pull request just for that effort.

>Adding the nav-paging buttons below and atop the table was a site-wide change we tested before. Would you prefer the fix to be scoped to the medications table only? or site-wide?

I think sitewide makes sense for similarly paginated index views.

Thanks!

tangollama:
@EvanSays any chance you're freed up to close this out now?

PR #1183, opened 2017-09-06
Update authenticate-user.js

fooksupachai:
Would it be better?

Fixes #[replace brackets with the issue number that your pull request addresses].

- [list out summary of changes here]
- [list out summary of changes here]
- [list out summary of changes here]
- [etc]

*Note: pull requests without proper descriptions may simply be closed without further discussion. We appreciate your contributions, but need to know what you are offering in clearly described format. Thanks! (you can delete this text)*

### DIFF
--- a/tests/helpers/authenticate-user.js
+++ b/tests/helpers/authenticate-user.js
@@ -2,9 +2,7 @@ import Ember from 'ember';
 import { authenticateSession } from 'hospitalrun/tests/helpers/ember-simple-auth';
 import { invalidateSession } from 'hospitalrun/tests/helpers/ember-simple-auth';
 
-const {
-  merge
-} = Ember;
+var { merge } = Ember ;
 
 Ember.Test.registerHelper('authenticateUser', function(app, attrs = {}) {
   let expiresAt = new Date().getTime() + 600000;


jkleinsc:
@fooksupachai if you run `ember test` locally on your changes, you will see the following error: 
```
not ok 250 Chrome 56.0 - ESLint - helpers/authenticate-user.js: should pass ESLint
    ---
        actual: >
            false
        expected: >
            true
        stack: >
                at Object.<anonymous> (serve://dist/assets/tests.js:11392:12)
        message: >
            helpers/authenticate-user.js should pass ESLint.
            5:1  - Always use `const` when making variables from Ember properties (ember-suave/require-const-for-ember-properties)
            5:1  - Unexpected var, use let or const instead. (no-var)
            5:23  - Unexpected whitespace before semicolon. (semi-spacing)
        Log: |
```

The reason for this is that the changes you made do not comply with the style guides HospitalRun uses:
https://github.com/dockyard/styleguides/blob/master/engineering/ember.md
and 
https://github.com/dockyard/styleguides/blob/master/engineering/javascript.md

This change reverses those style changes explicitly chosen and therefore I am going to close it.

PR #1211, opened 2017-09-15
Slack mention etiquette

WeeSkilz:
This pull request doesn't fix an issue explicitly, but addresses a conversation had in Slack over Slack etiquette from here https://hospitalrun.slack.com/archives/C02BT7ZBK/p1505380410000243

This updates the advice about what commands can and should be used, with the usage of `@everyone` and `@channel` being discouraged, and `@here` being deemed acceptable as it only notifies people currently active on Slack, as per https://get.slack.help/hc/en-us/articles/202009646-Make-an-announcement

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -10,6 +10,10 @@ However, before you ask in Slack "what can I contribute to", be sure to keep rea
 
 Also, please avoid use of the `@everyone` and `@channel` commands in Slack, as you will be sending a notification to nearly 600 people. Just post your question and someone will respond soon.
 
+While `@here` is discouraged as it notifies everyone who is active on Slack, if you have an announcement that the channel needs to hear urgently, use can be justified.
+
+Generally, just posting your question will allow you to recieve a timely answer.
+
 ## Help Wanted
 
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -8,7 +8,7 @@ Project communication occurs primarily and intentionally via our project [Slack]
 
 However, before you ask in Slack "what can I contribute to", be sure to keep reading this document for the answer to your question. :-)
 
-Also, please avoid use of the `@here` command in Slack, as you will be sending a notification to nearly 600 people. Just post your question and someone will respond soon.
+Also, please avoid use of the `@everyone` and `@channel` commands in Slack, as you will be sending a notification to nearly 600 people. Just post your question and someone will respond soon.
```
jglovier: This looks good, but maybe also include a note about `@here` something like:

> While `@here` will notify only people who are in the room and active, it's best to use this only if you have an announcement that everyone in the channel really needs to here.

WeeSkilz: Sure!

jglovier:
Thank again, @WeeSkilz!!
thumbs-up-daft-punk

WeeSkilz:
Always a pleasure!

Moana you're welcome

PR #1244, opened 2017-10-03
Add possible fix for `bower install` when root to README

BogdanAlexandru:
Fixes self-reported issue.

I run and always have into this issue with a clean install of Ubuntu and following the instructions, which I think is a common setup. I always have to take the mentioned step to get `bootstrap` to execute successfully.

I think that adding this to the README will save others' time, in the absence of a more permanent fix.

As for a more permanent fix:
How about taking `bower install` out of the bootstrapping script if we're not okay with permanently adding the `allow-root` flag?

<hr/>


tangollama:
Thx @BogdanAlexandru

PR #1276, opened 2017-10-21
Fix missing language error for language dropdown menu in navigation bar

weijentu:
Fix missing language error for language dropdown menu in navigation bar

Fixes #1276 .

- translations.js for each language under locales folder


BogdanAlexandru:
I also noticed this just now, was going to submit a PR but you were faster.

Could you also rename the `hindi` locale to `hi` (including the folder) so that it's consistent with the rest of the locales? I think that would help.

weijentu:
No problem, just added a commit 9f95180 according to your suggestion :)

Issue #1245, opened 2017-10-03
About the way translated language names are handled


BogdanAlexandru:
While working on the Romanian translation I stumbled upon this:

```js
// the French localization
languages: {
    en: 'Anglais',
    fr: 'Français',
    es: 'Espagnol',
    de: 'Allemand',
    // ...
  }
```

Each localization specifies its own language-specific names for the other languages. If one entry is missing, it shows up in the app as an alternative, "missing language" string.

But my question is rather about whether or not this approach helps the user. In all of the programs I've used, language names are not current-language specific. So for the French translation above, the fields would look like:

```js
// the French localization
languages: {
    en: 'English'
    fr: 'Français',
    es: 'Español',
    de: 'Deutsche',
    ro: 'Română',
    // and so on
  }
```

This is useful because if the program runs in a language that the user does not understand (or he accidentally picks an incorrect language), he can easily find his own language in the list (since it uses a familiar name). For some languages, the difference is quite large e.g. for French: "Allemand" vs "Deutsche", so I can see there being some difficulty. For other languages, the difference can be quite VERY LARGE. For example, when trying to switch to Romanian, I accidentally chose Urdu. I have no idea which one of these is Romanian, or even English, so I could go back:

![Imgur](https://i.imgur.com/FzdYkll.png)

While at it, would something like this kind of language selector not be more intuitive?
![img](https://alistapart.com/d/Column-Antoine/F1_blabla.png)
I can much quicker spot the flag of my country/other familiar country (language) in such a list, compared to a plain text list. Not to mention, it also looks like great eyecandy for very little effort.

Taking it out of the settings dropdown and displaying the flag directly would also remove the requirement for the current user to understand the currently selected language in order to tell where the language dropdown lies. This is not *that* big of an issue, especially given the really simple dropdown UI that HR+ currently has, but it's still some text sitting between the user and his preferred language, that he may not be able to understand.

![Imgur](https://i.imgur.com/VD4GJNU.png)

Flags shown directly are pretty universal and would ameliorate this issue.

Thoughts?

BogdanAlexandru:
[...]

PR #1278, opened 2017-10-23
Make language names in dropdown the same regardless of locale

BogdanAlexandru:
Fixes part of #1245

- Make language names in dropdown the same regardless of locale. This is useful so that users can find their preferred language easily regardless of what is already chosen.

### DIFF
--- a/app/components/language-select.js
+++ b/app/components/language-select.js
@@ -6,12 +6,18 @@ export default Ember.Component.extend({
 
   languageOptions: function() {
     let i18n = this.get('i18n');
-    return i18n.get('locales').map((item) => {
+    // Hacking around the fact that i18n
+    // has no support for t(key, locale).
+    let currentLocale = i18n.get('locale');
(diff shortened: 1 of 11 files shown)


BogdanAlexandru:
See it live here:

Imgur


This removes the need for each translation to specify the name of all other supported languages.
Each translation has its own `languageName` field at the top. This has a number of advantages:

* When a new translation is added, the other existing translation files do not need to be changed.
* There is no duplication involved.
* Less code to maintain.

Displaying the language options became a bit of a hack, due to the fact that ember-i18n does not
allow for one-shot custom locale translations, or any other natural way of specifying language names.
From what I saw on their repo, support for one-shot translations is going to be added but not sure when.
The hack can be removed once support is added.

In any case I opened up a question on ember-i18n's repo. We can maybe wait for an answer before merging,
maybe I missed something.

Kingmatrix:
New here anyone to help me out

PR #1295, opened 2017-11-15
address #1294: fix DOB unit tests

### LINKED ISSUE #1294: DOB formatting unit tests fail
**Expected behavior:**
All tests pass

**Actual behavior:**
2 tests fail (`convertDOBToText date string` and `convertDOBToText date string short format`) with the 1d difference

**Steps to reproduce:**
* Run tests

**OS and Browser:**
Ubuntu, Chrome

Some thoughts regarding this issue. 
1. I've fixed it (the PR is attached)
2. There was an error in the `dob` module when we're trying to parse a string. No one noticed that since this functionality is not used nowhere except unit tests. The fix (and the code) smells since it implies that we receive the date string in the specific format. _My question then: should we maybe remove the whole parsing part?_
3. I rolled back the #1268, because the author has fixed one test in the CST but broke 2 in the EST. The reason was that in the test we were passing the value 789109200000, which is
`new Date(789109200000) --> Tue Jan 03 1995 06:00:00 GMT+0100 (CET)` so it was the time part which broke the math here. I'd say that since we're working with dates only (the only consumer is the `patient` module which doesn't have a time component) we should work in the the same timezone. Or, let's convert everything to UTC then? For additional security I've put both dates there to the midnight. @aivong could you please verify that it works in your TZ?

If you have any thoughts regarding this, welcome :)

stukalin:
fix DOB unit tests

Fixes #1294.

- fix and improving readability of DOB formatting unit tests
- fix date parsing

### DIFF
--- a/app/mixins/dob-days.js
+++ b/app/mixins/dob-days.js
@@ -9,8 +9,12 @@ export default Ember.Mixin.create({
 
     if (birthDate) {
       if (birthDate.getFullYear === undefined) {
-        birthDate = moment(birthDate, 'l').toDate();
+        birthDate = moment(birthDate, 'LLL').toDate();
       }
+
+      birthDate.setHours(0, 0, 0, 0);
+      today.setHours(0, 0, 0, 0);
+
       if (birthDate.getFullYear !== undefined) {
         years = today.getFullYear() - birthDate.getFullYear();
         if (today.getMonth() < birthDate.getMonth()
@@ -37,7 +41,7 @@ export default Ember.Mixin.create({
       }
 
       if (birthDate.getDate) {
-        days = today.getUTCDate() - birthDate.getUTCDate();
+        days = today.getDate() - birthDate.getDate();
         if (days < 0) {
           days += 30;
         }
--- a/tests/unit/mixins/dob-days-test.js
+++ b/tests/unit/mixins/dob-days-test.js
@@ -4,6 +4,7 @@ import Ember from 'ember';
 import DS from 'ember-data';
 import tHelper from 'ember-i18n/helper';
 import localeConfig from 'ember-i18n/config/en';
+import moment from 'moment';
 
 moduleFor('mixin:dob-days', 'Unit | Mixin | dob-days', {
   needs: [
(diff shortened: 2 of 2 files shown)


app/mixins/dob-days.js
```diff
@@ -9,8 +9,12 @@ export default Ember.Mixin.create({
 
     if (birthDate) {
       if (birthDate.getFullYear === undefined) {
-        birthDate = moment(birthDate, 'l').toDate();
+        birthDate = moment(birthDate, 'LLL').toDate();
```
stukalin: I think this if should be removed (see issue)

tests/unit/mixins/dob-days-test.js
```diff
@@ -47,26 +48,32 @@ moduleFor('mixin:dob-days', 'Unit | Mixin | dob-days', {
   }
 });
 
-test('convertDOBToText', function(assert) {
+test('convertDOBToText date object', function(assert) {
   let dobDays = this.subject();
 
-  assert.strictEqual(dobDays.convertDOBToText(new Date(789109200000)).toString(), '21 years 11 months 12 days');
+  assert.strictEqual(dobDays.convertDOBToText(moment('January 3rd, 1995', 'LLL').toDate()).toString(), '21 years 11 months 12 days');
```
stukalin: better readability with moment.js

tests/unit/mixins/dob-days-test.js
```diff
@@ -47,26 +48,32 @@ moduleFor('mixin:dob-days', 'Unit | Mixin | dob-days', {
...
+  let dobDays = this.subject();
+
+  assert.strictEqual(dobDays.convertDOBToText(moment('January 3rd, 1995 6:00 AM', 'LLL').toDate()).toString(), '21 years 11 months 12 days');
 });
 
 test('convertDOBToText date string', function(assert) {
   let dobDays = this.subject();
 
-  assert.strictEqual(dobDays.convertDOBToText('January 3rd, 1995').toString(), '21 years 8 months 26 days');
+  assert.strictEqual(dobDays.convertDOBToText('January 3rd, 1995').toString(), '21 years 11 months 12 days');
```
stukalin: It was the same date

PR #1306, opened 2017-11-22
posibility to add documents

### LINKED ISSUE #1286: Upload Documents
**Expected behavior:** Possibility to upload and view pdfs. In order to can track a patients cases, I think it would be useful to add the possibility to upload documents as pdf.

**Actual behavior:** only photos

**Steps to reproduce:** In the same way that we have the photos menu, have one with documents

**OS and Browser:**

egibar:
Fixes #1286.

Possibility to upload and view pdfs to the page.
I not sure but I thinks that only uploads in local, no to the DB, i think it could be because of the db-listeners.


egibar:
Now I am too busy so I can't do it, also I am not very familiar with the test and how to do it, so I will appreciate if someone can help me. Other point, I suppose that we have to change the dblistenners in order to upload the documents to couchDB

PR #1318, opened 2017-11-30
Docker fix

stukalin:
- fix Dockerfile since alpine doesn't have many required packages
- enable autoreloading via docker-compose
- readme fix

Somehow #1300 brought issues in Docker container building which were not identified right away (very-very strange). Proposed is the fix for that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,34 @@
 FROM node:6-alpine
 
 # install script dependencies
-RUN apk update && apk add sudo curl git wget
+RUN apk update && apk add \
+  bash \
+  curl \
+  g++ \
+  git \
+  make \
+  python \
+  sudo \
+  wget
+
+# install global npm dependencies
+RUN npm install -g ember-cli@latest && npm install -g bower
+
+# use changes to package.json to force Docker not to use the cache
+# when we change our application's nodejs dependencies:
+COPY package*.json /tmp/
+RUN cd /tmp && npm install
+RUN mkdir -p /usr/src/app && cp -a /tmp/node_modules /usr/src/app
+
+COPY bower.json /tmp/
+RUN cd /tmp && bower install --allow-root
+RUN cp -a /tmp/bower_components /usr/src/app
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,20 +1,34 @@
...
+  curl \
+  g++ \
+  git \
+  make \
+  python \
+  sudo \
+  wget
+
+# install global npm dependencies
+RUN npm install -g ember-cli@latest && npm install -g bower
```
stukalin: Here I've slightly changes the order of commands in order to use docker cache

Dockerfile
```diff
@@ -1,20 +1,34 @@
...
+
+# use changes to package.json to force Docker not to use the cache
+# when we change our application's nodejs dependencies:
+COPY package*.json /tmp/
+RUN cd /tmp && npm install
+RUN mkdir -p /usr/src/app && cp -a /tmp/node_modules /usr/src/app
+
+COPY bower.json /tmp/
```
stukalin: bower goes here

README.md
```diff
@@ -80,7 +80,7 @@ To run HospitalRun with [Docker](https://www.docker.com/) please do the followin
 - Clone the repository with the command `git clone https://github.com/HospitalRun/hospitalrun-frontend.git`.
 - Change to the hospitalrun-frontend directory `cd hospitalrun-frontend`.
 - Build the HospitalRun image with `docker build -t hospitalrun-frontend .`.
-- Execute `docker run -it --name couchdb -d couchdb` to create the couchdb container.
+- Execute `docker run -it --name couchdb -d couchdb:1.7.1` to create the couchdb container.
```
stukalin: Couchdb 1.* is our primary

Dockerfile
DerekTBrown: you don't need this

stukalin: Why not? Saves several seconds time on rebuild. I remember that we're gonna get rid of bower but we haven't yet

PR #968, opened 2017-02-27
added the ability to edit role names

bradgearon:
Fixes #935 

**Changes proposed in this pull request:**
- added input box for role name to admin/roles
- using role id for role select value in view
- using role element in selectRole action / storing selected role id
- added namedRoles map to user roles (filtered roles in user roles controller now uses this)
- updating loaded userRoles on save with new name (using unmodified role id)
- loading user roles and using role name on users index

image

cc @HospitalRun/core-maintainers

PR #1320, opened 2017-12-01
Change role names

DerekTBrown:
Continuation of https://github.com/HospitalRun/hospitalrun-frontend/pull/968


DerekTBrown:
@tangollama I am working on this right now.  The problem it seems is that this would require stubbing the database further for many of the unit tests.

tangollama:
@DerekTBrown Hello. I'm pulsing just to determine if we should leave this PR open, close it, or perhaps reach of to the community for someone else to finish it. Can you comment?

Issue #1156, opened 2017-09-01
Moving away from bower?


tangollama:
**Expected behavior:**
npm install with no warning

**Actual behavior:**
npm WARN deprecated bower@1.8.0: ..psst! While Bower is maintained, we recommend Yarn and Webpack for *new* front-end projects! Yarn's advantage is security and reliability, and Webpack's is support for both CommonJS and AMD projects. Currently there's no migration path, but please help to create it: https://github.com/bower/bower/issues/2467

**Steps to reproduce:**
checkout the project
navigate to the frontend directory
npm install

**Screenshots (if applicable):**
N/A

**OS and Browser:**

DerekTBrown:
Probably should consider moving to yarn instead.  Working on a PR.

stukalin:
Why `yarn` und not simply `npm`? It has package.lock now etc.

DerekTBrown:
@stukalin I think you may be right.  Per your suggestion in the PR, I am going to try to break this up into smaller pieces.  Working on a PR which just moves everything into NPM.

DerekTBrown:
In order to facilitate this, we are going to need to add a `package.json` file into the hospitalrun-maintained fork of `JsBarcode`. Opening a new PR.

stukalin:
Could be useful. Things which are to be migrated 
- [ ] JsBarcode
- [ ] bootstrap 
- [ ] filer.js
- [ ] idb.filesystem
- [ ] octicons
- [ ] pikaday
- [ ] typeahead.js
- [ ] pouchdb-load
- [ ] pretender 
- [ ] fullcalendar 
- [ ] fullcalendar-scheduler
- [ ] timekeeper
- [ ] webrtc-adapter

donaldwasserman:
I'll just close this one.

PR #1352, opened 2018-01-25
ember 2.11 from npm

stukalin:
address #1156.

- remove ember from bower and run it from npm
- to start with I use v2.11
- mandatory update of `ember-rapid-forms`
- corresponding update of all the forms
- tests fix
- package-lock.json (can be discussed though)

Some comments:

1. I used v2.11 because it was the next version to v2.10, so the number of deprecations is minimum. Afterwards we could jump directly to v2.18 or whatever
2. Unfortunately `ember-rapid-forms` must have been updated because of its dependencies. And the sad thing about that was that between `beta10` and `beta11` so many changes were made that it would have been impossible to keep the existing code. I have updated to the latest version and tried to fix all the occurencies. Tests helped me here.
3. Could be related to the previous one, but now we have to drag model in our custom components. That's why there are all those `model=model`

I'll put some more comments in the code.


app/admin/address/template.hbs
```diff
@@ -1,15 +1,15 @@
...
-      {{#em-form model=model submitButton=false }}
-        {{em-input label=(t 'admin.address.address1Label') property="value.address1Label"}}
-        {{em-checkbox label=(t 'admin.address.include1Label') property="value.address1Include"}}
-        {{em-input label=(t 'admin.address.address2Label') property="value.address2Label"}}
-        {{em-checkbox label=(t 'admin.address.include2Label') property="value.address2Include"}}
-        {{em-input label=(t 'admin.address.address3Label') property="value.address3Label"}}
-        {{em-checkbox label=(t 'admin.address.include3Label') property="value.address3Include"}}
-        {{em-input label=(t 'admin.address.address4Label') property="value.address4Label"}}
-        {{em-checkbox label=(t 'admin.address.include4Label') property="value.address4Include"}}
+      {{#em-form model=model submitButton=false as |form|}}
```
stukalin: This is how those rapid forms work now.

app/admin/custom-forms/edit/template.hbs
```diff
@@ -61,8 +61,8 @@
         <h3 class="panel-title">{{formName}}</h3>
       </div>
       <div class="panel-body">
-        {{#em-form model=previewModel submitButton=false}}
-          {{custom-form model=previewModel form=model}}
+        {{#em-form model=previewModel submitButton=false as |form|}}
+          {{custom-form model=previewModel form=model rapidForm=form}}
```
stukalin: Here we use the `form` ourselves, so I pass rapid form as `rapidForm`. Maybe it makes more sense to do `rapidForm=form` everywhere. Opinions?

app/admin/lookup/route.js
```diff
@@ -12,10 +12,6 @@ export default AbstractIndexRoute.extend({
     return this.store.findAll('lookup').catch((error) => this.send('error', error));
   },
 
-  afterModel(model) {
```
stukalin: em-select can choose  the first element automatically now, but I did that only here when I realised how big this PR would be

app/appointments/calendar/route.js
```diff
@@ -76,10 +76,10 @@ export default AppointmentIndexRoute.extend({
         });
       }
     });
-    return {
+    return Ember.Object.create({
```
stukalin: That adds all those get/set methods which are now expected

app/appointments/edit/template.hbs
```diff
@@ -1,27 +1,27 @@
...
-      {{patient-typeahead property="patientTypeAhead" label=(t 'models.patient.names.singular') content=patientList selection=selectedPatient class="required test-patient-input"}}
+      {{patient-typeahead model=model property="patientTypeAhead" label=(t 'models.patient.names.singular') content=patientList selection=selectedPatient class="required test-patient-input"}}
     {{else}}
       {{patient-summary patient=model.patient returnTo='appointments.edit' returnToContext=model.id disablePatientLink=model.isNew }}
     {{/if}}
     <div class="row">
       {{#if isAdmissionAppointment}}
-        {{date-picker property="startDate" label=(t 'models.appointment.labels.startDate') showTime=showTime class="col-sm-4 required test-appointment-start"}}
-        {{date-picker property="endDate" label=(t 'models.appointment.labels.endDate') showTime=showTime class="col-sm-4 required test-appointment-end"}}
+        {{date-picker model=model originalPropery="startDate" property="display_startDate" label=(t 'models.appointment.labels.startDate') showTime=showTime class="col-sm-4 required test-appointment-start"}}
```
stukalin: with date pickers it became not so nice: now we should pass 2 properties with the correct signature.... I'd try to fix this somehow. Will see.

app/components/date-picker.js
```diff
@@ -1,19 +1,11 @@
...
   dateSetAction: null, // Specify an action to fire when a date is selected
-  htmlComponent: 'date-input',
   minDate: null,
   maxDate: null,
   originalPropery: null,
   showTime: false,
-  yearRange: 10,
-
-  didReceiveAttrs(attrs) {
-    let dateProperty = this.get('property');
```
stukalin: I had to remove this because Ember now tracks if some property has been set more than once and throws an error.

PR #1375, opened 2018-02-19
Analytics

### LINKED ISSUE #115: Add analytics
It may be too early to think much about this, but something that will be really helpful for improving how people use the app is installing some analytics to see what pages and routes users are using most, and gain UX insights from. The Atom team uses Google Analytics and a [custom package](https://github.com/atom/metrics) for this, which also has the benefit of providing aggregate stats about how many total users there are for the app, and what their activity on a weekly/monthly basis is like.

stukalin:
Fixes #115 .

- Google Analytics via `ember-metrics` library
- offline mode support with a lib from Google

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -1,9 +1,28 @@
 import EmberRouter from '@ember/routing/router';
 import config from './config/environment';
+import { get } from '@ember/object';
+import { inject as service } from '@ember/service';
+import { scheduleOnce } from '@ember/runloop';
 
 const Router = EmberRouter.extend({
+  metrics: service(),
+
   location: config.locationType,
-  rootURL: config.rootURL
+  rootURL: config.rootURL,
+
+  didTransition() {
+    this._super(...arguments);
+    this._trackPage();
+  },
+
+  _trackPage() {
+    scheduleOnce('afterRender', this, () => {
+      let page = this.get('url');
+      let title = this.getWithDefault('currentRouteName', 'unknown');
+
+      get(this, 'metrics').trackPage({ page, title });
+    });
+  }
 });
 
 Router.map(function() {
--- a/app/router.js
+++ b/app/router.js
@@ -4,6 +4,9 @@ import { get } from '@ember/object';
 import { inject as service } from '@ember/service';
 import { scheduleOnce } from '@ember/runloop';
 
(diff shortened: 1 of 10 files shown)


config/environment.js
```diff
@@ -7,6 +7,23 @@ module.exports = function(environment) {
     environment: environment,
     rootURL: process.env.EMBER_CLI_ELECTRON ? null : '/',
     locationType: 'hash', // Using hash location type because it is more friendly for offline.
+    metricsAdapters: [
+      {
+        name: 'GoogleAnalytics',
+        environments: ['development', 'production'],
+        config: {
+          id: 'UA-TODO-HOSPITAL-RUN',
```
stukalin: @tangollama we'd need the ID here

package.json
```diff
@@ -114,8 +115,8 @@
   "dependencies": {
     "electron-compile": "^6.4.2",
     "electron-localshortcut": "^3.0.4",
-    "electron-protocol-serve": "1.3.0",
```
stukalin: that was duplicated somehow

app/router.js
```diff
@@ -1,9 +1,28 @@
...
+    this._super(...arguments);
+    this._trackPage();
+  },
+
+  _trackPage() {
+    scheduleOnce('afterRender', this, () => {
+      let page = this.get('url');
+      let title = this.getWithDefault('currentRouteName', 'unknown');
+
+      get(this, 'metrics').trackPage({ page, title });
```
stukalin: Well, this piece of code was copy-pasted from the `ember-metrics` sample. Apparently we'd need something more sophisticated

stukalin:
@jglovier it's surely the former and yes, you're right, it's confusing now which GA account should be used. It also allows everyone to change the config easily, so maybe some encapsulation of GA-ID would be required there.

stukalin:
So, I see 3 solutions there:
1. If we really say "Oh, no one really edits `config/environment` on their own!", let's leave it as it is.
2. We would be able to use [Lazy initialization](https://github.com/poteto/ember-metrics#lazy-initialization) in order to encapsulate _a bit_ the ID and put it in our `route.js`.
3. And if we really care that our ID is... well... call it safe or maybe not-so-confusing, then the possible solution could be creating a custom `hr-google-analytics` metrics adapter and encapsulate the ID there.

@jglovier @tangollama @jkleinsc opinions?

jglovier:
@stukalin thanks for the response. @jkleinsc @tangollama I'd be interested to hear your thoughts on ☝️

Issue #1382, opened 2018-02-28
Replace ember-validations


internets:
```ember-validations``` is no longer under active development, and has become a blocker for completion of #1370 (ember-getowner-polyfill.import deprecation). There's a PR that's over a year old that is unlikely to ever be accepted: DockYard/ember-validations#442. 

The way I see it there are two options: fork or replace.

If fork, it's possible that forking and cherry-picking the commit from that PR will solve the immediate problem.

If replace, the two most common actively developed validation solutions to my knowledge are ember-cp-validations and ember-changeset-validations. I recommend ember-cp-validations for the following reasons:

- The broad strokes of the API are very similar to ember-validations
- It doesn't require buying into the ember-changeset way of handling form objects

broerse:
@martndemus can you comment on what state `ember-validations` is in. Is it on the DockYard roadmap to update it to Ember 3.0?

martndemus:
We are not planning to do anything about ember-validations.

stukalin:
Well, it seems to be a large piece of work, migrating it. I've started the new branch but the process is not so straightforward. See https://github.com/HospitalRun/hospitalrun-frontend/compare/ember-validations?expand=1

The most important problems are:
- having all validations to be transferred correctly. It is not clear copy-paste unfortunately even though some settings in both libs match
- conditional validations are the problem as well: our current solution supports some "hacks" in order to make validation conditional. For instance have an acceptance validator with an `if` function. I tried to get rid of that and use a custom validator.
- I believe we've lost i18n support. So this part should be verified as well.

To sum up, maybe forking would be a good quick solution.

donaldwasserman:
@stukalin - What would be the best approach to forking this? Should someone who have repository creation permissions fork it to the Hospital Run Repo? It's probably better than running it off someone's personal fork.

stukalin:
That'd be the best option. I don't have sufficient right, I've just verified that. @jglovier, could you help forking the https://github.com/DockYard/ember-validations to our repo?

stukalin:
So now it's only matter of applying the fix mentioned in the very first message to our fork and reference it from the main app.

donaldwasserman:
I'm going to try to get that done this morning (well, morning East Coast US).

Issue #1370, opened 2018-02-14
Fix deprecations from deprecation-workflow (and probably remove it at all)


stukalin:
We have a plugin `ember-cli-deprecation-workflow` which allows us to hide some deprecation messages. However if we want to be able to update Ember to the v3 we'd better fix those deprecations. 

Some of them are brought the project dependencies, so if there aren't new version of those dependencies let's at least identify them at list them down. 

If all deprecations will be fixed I suggest removing the plugin at all.

- [ ] ember-getowner-polyfill.import (should be fixed in #1382)
- [x] ember-metal.binding
- [x] ember-views.did-init-attrs (should be fixed in #1369)
- [x] ds.serializer.private-should-serialize-has-many (should be fixed with the `ember-pouch` update)

FrankieDickinson:
I will attempt to fix this. First time contributor.

tangollama:
Thx. 

Typed or dictated imperfectly on my iPhone (717.385.9970)

internets:
@stukalin I've been doing a lot of 2.18 updates for 2.x ember apps lately, I'd like to take this on. Ultimately I'd like to take on #1365 as well, but doing this first should make the PRs more manageable. Would you be open to marking #1365 as in progress as well to prevent some potentially ugly merge scenarios?

It looks like #1369 and #1366 are in progress and will be addressing some aspects of this PR. If those turn out to be non-trivial, I can wait for them to land on master and rebase.

stukalin:
@internets thanks for taking this over. I marked these issues as `in progress`. Dunno why I can't assign them on you.

internets:
No problem, happy to contribute! 

I believe you can't assign them to me because github requires me to have push access to the repo. Not a popular limitation :) The ```in progress``` label should be enough, thanks!

alexanderbrown21:
@internets feel free to work on #1365 :). It's worth someone with experience of doing existing migration to touch. i'm afraid don't have much free time as can at the moment.

internets:
@alexanderbrown21 thanks!

@stukalin I've made some progress which can be viewed here: https://github.com/HospitalRun/hospitalrun-frontend/compare/master...internets:1370

Before I open a PR I wanted to go over some notes and questions.

### ember-getowner-polyfill.import

[...]

Issue #1369, opened 2018-02-14
Replace ember-select-list with em-select


stukalin:
The `ember-select-list` (https://github.com/begedin/ember-select-list) was last modified in 2015. We use another library - `ember-rapid-forms` which has select helper as well (http://piceatech.github.io/ember-rapid-forms/#/controls/select). So I believe `ember-select-list` could be removed completely in favor of rapid-forms.

Additionally we'll get rid of a deprecation message `ember-views.did-init-attrs`. So, whoever takes this please remove the `deprecation-workflow.js` record.

stukalin:
@ben-muldrow any luck with the issue?

msalahz:
Hi, I'm available to take care of this, Is this in progress or I can take it ?

ben-muldrow:
@stukalin Still hammering away! Been swamped lately -- going to try to finish by the start of next week.

efx:
bumping the issue as I'm looking to contribute.
@ben-muldrow could you link your branch of what you have so far?

ben-muldrow:
@efx Sure thing! Sorry it's taking so long, I got stuck on one of the instances. Would love to go over it with you if you have time. Here's my branch with progress so far https://github.com/ben-muldrow/hospitalrun-frontend/tree/ember-select-list-deprecation

efx:
Sweet, no problem at all. Fixing deprecations in ember apps can be quite grueling. I'll peek through that branch later today or tomorow. Perhaps sometime in the next couple weeks we could connect over Hospital Run's slack if you want to go over it together.

MatthewDorner:
I'd be able to join the discussion as well. I looked at it a little bit before.

MatthewDorner:
@efx @ben-muldrow Any updates on this? I started on it myself and got it replaced on all instances and working, just need to clean up a little then I'll post my code. I'd like to discuss it though.

MatthewDorner:
ok, I started a conversation on there

efx:
We are actively working on this in this branch: https://github.com/MatthewDorner/hospitalrun-frontend/tree/em-select-squashed. I'll be giving PR reviews on those changes until we can open a single squashed PR against Hospital Run master.

PR #1383, opened 2018-02-28
Fix deprecations from deprecation-workflow

internets:
Addresses #1370.

- Update ```ember-concurrency```, ```ember-i18n```, ```ember-pouch```, ```ember-simple-auth```, ```ember-sinon-qunit```, ```worker-pouch```
- Replace Ember.Binding calls with Ember.defineProperty calls
- Fix backtracking rerender error in inventory-location-picker.js (caused by replacing Ember.Binding calls)
- Replace stub(obj, meth, stubFn) with stub(obj, meth).callsFake(stubFn) in sinon tests

(note that this does not completely address #1370. #1369 and #1382 must be completed before ```ember-cli-deprecation-workflow can be removed``` and  #1370 can be closed)


internets:
A-ha, I did not run the electron tests. I'll get those passing.

stukalin:
@internets do you need any help with electron tests?

internets:
@stukalin Yes! Sorry for going radio silent for so long, I've been trying to carve out more time to address this. Both this codebase and electron debugging are new to me so it's been time consuming. Want me to catch you on slack to discuss, or do you want to pull and take a stab at it?

P.S. Yes we can remove those deprecation-workflow config entries, commit incoming.

stukalin:
@internets I've found the reason. The `i18n` library contains some incompatible change in the version 5.1. Rolling it back to 5.0 fixed some of the tests. We'll see what Travis says:)

internets:
Interesting! I tried that downgrade and it did not solve the problem using yarn, but it does work with npm. I'll see if I can figure out the difference. Thanks!

internets:
@stukalin So far my investigations show that even with the same installed versions, using yarn to install results in test failures and using npm results test success. I must be missing something! I'll leave it to you whether you want to merge it the way it is. The electron app does not exhibit the bugs indicated by the test failures when installing via yarn, it only happens in the test environment. 

I don't like giving up on things so I'll continue investigating, but I don't know how long it will take me. I can open a second PR when I figure it out if need be.

Issue #1401, opened 2018-03-26
Highlight current subroute in `sidebar-nav`


FalkF:
I might have found my personal `good first issue` here :)

While familiarizing myself with the app, I missed highlighting of the active subroute in the `sidebar-nav`.
I'm thinking of icon indentation and color, similar to the hoover effect. 

I would like to make a PR if anyone agrees.

**Screenshots:**
on route `patients/admitted`
![hr-highlighting1](https://user-images.githubusercontent.com/19377375/37921514-53960058-312a-11e8-98cc-fde50c619ae8.png)
hoover effect
![hr-highlighting2](https://user-images.githubusercontent.com/19377375/37921516-55751422-312a-11e8-8dfb-9dcb4d6125a7.png)

PR #1403, opened 2018-03-26
1401 highlight current route

FalkF:
Fixes `nav-sidebar` highlighting of sub and main routes.

Before: #1401 
Now:
hr-highlighting3

Closes: #1401 

### DIFF
--- a/app/mixins/navigation.js
+++ b/app/mixins/navigation.js
@@ -12,7 +12,7 @@ export default Ember.Mixin.create({
     {
       title: 'Inventory',
       iconClass: 'octicon-package',
-      route: 'inventory.index',
+      route: 'inventory',
       capability: 'inventory',
       subnav: [
         {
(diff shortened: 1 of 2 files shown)


FalkF:
If we want the `link-to` helper of the parent route to be active while visiting a subroute, the parent route needs to be part of the subroute:
`inventory.listing` contains the parent `inventory`, so the parent and child links are set to active.

The standard subroute to open, when visiting a parent route should be defined with `{ path: '/' }` in the router afaik.
(e.g. for the [admin route](https://github.com/HospitalRun/hospitalrun-frontend/blob/c68eaec8b212901c96e4a0041db6f30c462d1230/app/router.js#L22)

FalkF:
@tangollama does it make sense, or do I need to make adjustments?

tangollama:
Thanks @FalkF for the response. Makes sense.

Issue #1366, opened 2018-02-10
Move 'navigation' to a component in order to fix the deprecation


stukalin:
**Expected behavior:**
No deprecation warning

**Actual behavior:**
There is a warning

> DEPRECATION: Please refactor `{{render "navigation"}}` to a component and invoke via `{{navigation}}`. (\\\'hospitalrun/templates/application.hbs\\\' @ L2:C2)  [deprecation id: ember-template-compiler.deprecate-render] See http://emberjs.com/deprecations/v2.x#toc_code-render-code-helper for more details.

**Steps to reproduce:**
Run `ember test` for instance

James-Byrne:
I can have a go at this if its available

stukalin:
It's available. Please, feel free to take

11 февр. 2018 г. 5:11 ПП пользователь "James Byrne" <
notifications@github.com> написал:

stukalin:
@James-Byrne any luck with the issue?

msalahz:
Hi, I'm available to take care of this, Is this in progress or I can take it ?

James-Byrne:
Hey sorry for the delay in replying, I got side tracked while working on it. @msalahz if you want to take it up go ahead.

msalahz:
Got it, Thanks @James-Byrne.
@tangollama: I'm taking over this issue.

msalahz:
Hi @tangollama , @stukalin , Sorry I will not be able to complete this task because I got a new job that made me very busy.
The work done on this task can be found [here](https://github.com/msalahz/hospitalrun-frontend/commits/1366-render-deprecation) in case anyone want to start from what I got done instead of starting from scratch.
Look forward to get back to you soon.

MatthewDorner:
Hey. I've been looking at this and made some progress, but I'm still learning the framework so I'm not too confident about it yet. I will look at your code though, and compare with what I have.

stukalin:
@msalahz @MatthewDorner I think I've finished the issue. See the attached PR and don't hesitate to review.

stukalin:
Oh, @msalahz , would you mind to sign the CLA? https://github.com/HospitalRun/hospitalrun-frontend/pull/1397#issuecomment-374610817

msalahz:
Done @stukalin , Thank you for completing my work.

PR #1406, opened 2018-03-28
Fix search bar text field

MatthewDorner:
Change in behavior was introduced in the fix for #1366: The text in the search bar used to get cleared if you 1) submitted a search or 2) moved to a different nav item. Now, the text remains in the search bar unless you clear it yourself.

Steps to reproduce:
- With current c68eaec code, perform a search and observe the text remains in the search field (top screenshot).
- Move to a different nav item and observe the text still stays there (bottom screenshot).
- Checkout commit 215c1fc from before #1366 fix and perform the same steps, observe the text is cleared in both cases (perform search, or move to different nav item).

search_screenshot

- pass searchText property into the sidebar-nav component to restore the original behavior.

** note: I observed the code to clear the search text is inside the search() action. the only reason it gets cleared when you move to a different nav item is because every time you move to a different nav item after having entered search text, it performs your search before it transitions. I guess this is because it's running that search action any time you blur the text field. my code doesn't change anything about this, but I could look at making it so it only searches on enter or clicking the search button if you think that would be worthwhile.


MatthewDorner:
cool, I should be able to add that test.

MatthewDorner:
Added the acceptance tests to the navigation-test.js file, does that seem like the right place for them? Also confirmed that tests fail if the line in application.hbs is removed, and pass once the line is added.

stukalin:
@MatthewDorner your changes were approved but some tests fail on Travis. Could you please fix them?

MatthewDorner:
oh, yeah. I see why it's failing now. strangely I didn't get those when I ran tests on my system. I will fix it.

MatthewDorner:
Sorry about that, had to change the solution slightly, but I think it's better this way anyway.

tangollama:
Thank you @MatthewDorner for the PR!

PR #1424, opened 2018-04-29
Fixes #1369 select-list deprecation

### LINKED ISSUE #1369: Replace ember-select-list with em-select

MatthewDorner:
Fixes #1369.

- The new ember-rapid-forms select helper mutates the specified property automatically, while the old one required an action. This is a reason for some of the refactoring.
- Other refactoring because the new one wouldn't work with a simple array (at least I couldn't get it to), so there's some code to change arrays into key/value object arrays.
- This PR will fail one of the acceptance tests until a new version of ember-rapid-forms is released which fixes an issue encountered in ember-rapid-forms during this work. This will probably be within the next couple weeks. See: https://github.com/piceaTech/ember-rapid-forms/issues/193
- In one instance, language-select.hbs, I went with a native HTML select instead of the ember-rapid-forms select, because the ember-rapid-forms one was adding Bootstrap CSS and caused the appearance to change. Let me know if you'd prefer it done another way. See imgur.com/a/h0ljX1A.
- Thanks @ben-muldrow for the initial work on this, @efx for reviewing and @gcorbel for helping us with ember-rapid-forms.


app/invoices/edit/template.hbs
```diff
@@ -135,14 +135,13 @@
                       </div>
                     </td>
                     <td class="col-xs-1">
-                      <div class="form-group">
-                        {{select-list
-                          class='form-control'
+                      {{#em-form model=detail submitButton=false as |form|}}
```
stukalin: Does it work well? Because you kinda introduce a form within a form.

MatthewDorner: It seems to work the same as before. One thing is you have to define "model=detail" inside the {{#em-form}}, and the form.select uses that to find its property. So I'm not sure if it can work if I define the {{#em-form}} outside of that {{#each item.details as |detail|}} block, though I can try it.

MatthewDorner:
Alright, the form-within-a-form is definitely something I should fix. After more research, it's not valid HTML, and also doesn't work on FireFox. So I will start looking into this and update in couple days.

app/invoices/edit/template.hbs
stukalin: alright. then I make the PR to "require changes"

MatthewDorner:
Fixing that also allowed me to remove the added computed property in invoices/edit controller. Now it should be ready to finish reviewing.

app/templates/components/location-select.hbs
```diff
@@ -1,8 +1,10 @@
-{{select-list
-  value=locationPicker.selectedLocation
-  content=locationPicker.subLocationList
-  class="form-control"
-  optionValuePath="selectObject"
-  optionLabelPath="selectObject.locationNameWithQuantity"
-  action=locationChange
-}}
\ No newline at end of file
+{{#em-form model=this submitButton=false as |form|}}
```
stukalin: Now I'm wondering if this doesn't introduce the same issue "form within a form" since we use that in a component which could possibly be used in a form. Could you please doublecheck?

stukalin: And sorry for the PR ping-pong

MatthewDorner: Yeah, it is used in a form. This one still works in Firefox unlike the first one, but it is still incorrect. Also quantity-conv.hbs does this, maybe others. Thanks for catching this, I just didn't know that about the HTML form tag. I'll have to think about how to handle this.

stukalin: Maybe there is already a rapid form which contains a component. In that case you should be able to pass the rapid form reference in the component and you're done.

MatthewDorner: Ok, will try that.

MatthewDorner: Some of these are multiple levels of component down from the {{#em-form}}, quantity-conv.hbs for example is inventory/edit/template -> (partial 'inv-purchase') -> quantity-calc -> quantity-conv, so it could end up a little confusing. Of course the alternative is another native HTML. It's up to your preference.

MatthewDorner: OK, it works but I had to also add "model=this" in the individual {{form.select}}'s.

stukalin:
@MatthewDorner while it's not merged... Would you mind to remove `ember-views.did-init-attrs` from our deprecation-workflow list as of #1370? It shouldn't throw this deprecation anymore.

tangollama:
Thank you @MatthewDorner for the PR!

PR #1430, opened 2018-05-13
Added ISO codes to language drop down

### LINKED ISSUE #1245: About the way translated language names are handled

davidackroyd99:
Fixes #1245

**In the language select drop down, you now get ISO code as well as language name:**
- Changed translations.js, now includes isoCode field
- language-select.js pulls that out as well as the language name
- language-select.hbs now displays it in dropdown, ie 'EN - English'

This is my first pull request and I'm very excited to be working on this project. I don't know if this is particularly helpful, but anything you want me to do, just say.

### DIFF
--- a/app/components/language-select.js
+++ b/app/components/language-select.js
@@ -17,7 +17,8 @@ export default Component.extend({
       i18n.set('locale', item);
       return {
         id: item,
-        name: i18n.t('languageName')
+        name: i18n.t('languageName'),
+        iso: i18n.t('isoCode')
       };
     });
     i18n.set('locale', currentLocale);
--- a/app/locales/cn/translations.js
+++ b/app/locales/cn/translations.js
@@ -1,5 +1,6 @@
 export default {
   languageName: '简体中文',
+  isoCode: 'CN',  
   admin: {
     address: {
       address1Label: '地址标签1',
--- a/app/locales/de/translations.js
+++ b/app/locales/de/translations.js
@@ -1,5 +1,6 @@
 export default {
   languageName: 'Deutsch',
+  isoCode: 'DE',
   admin: {
     address: {
       address1Label: 'Adresse 1 Kennzeichen',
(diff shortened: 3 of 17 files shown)


stukalin:
Note, that you have a number of tests to fix.

davidackroyd99:
I will get these tests fixed tonight and recommit.

stukalin:
@davidackroyd99 any luck with tests?

PR #1435, opened 2018-05-17
Turn search feature off for Incidents as per issue 1411

### LINKED ISSUE #1411: "Incident" displays search field but search doesn't work
**Expected behavior:**
Search control is not shown for pages that don't have search capability.

**Actual behavior:**
"Incident" shows the search field, but if you try to search it just hangs on the search dialog.

**Steps to reproduce:**
Log in as someone with access to "Incident".
Click "Incident" on the nav sidebar.
Observe that the search control is visible at the top of the nav sidebar.
Enter a search term and search.

**OS and Browser:**
Ubuntu 16.04 and Chromium.

Since I can't find see any code for a search route for Incident, I'm guessing Incident is not supposed to have a search and it just needs to have allowSearch set to false in the main Incident route so the search control is not displayed.

nzidol:
Fixes #1411.

- Quick solution is to add allowSearch: fasle, to app/incident/route.js
- Note added translated titles to route.js
- A feature request should be raised to add search functionality...

### DIFF
--- a/app/incident/route.js
+++ b/app/incident/route.js
@@ -10,6 +10,10 @@ export default AbstractModuleRoute.extend({
     name: 'incidentCategoryList',
     findArgs: ['inc-category']
   }],
+  allowSearch: false,
+  currentScreenTitle: t('incidents.titles.incidents'),
+  editTitle: t('incidents.titles.editIncident'),
+  newTitle: t('incidents.titles.newIncident'),
   moduleName: 'incident',
   newButtonText: t('incident.buttons.newIncident'),
   sectionTitle: t('incident.titles.incidents')


nzidol:
updated the file to pass the test, waiting for a rerun....

stukalin:
@nzidol thanks for the PR. I'd like first to hear opinions of the people who know exactly if we should disable the search feature for incidents or we should allow incidents to be searched. @tangollama @jkleinsc would you mind to look at the #1435 and decide what is correct?

nzidol:
My thinking is that it needs to be turned off until search has been
implemented.....

Cheers

Ido

PR #1441, opened 2018-05-22
Changed amountOwedChanged back to observer.

### LINKED ISSUE #1440: Invoice computed properties not updating.
**Expected behavior:**
Various totals and calculated values on "Edit Invoice" page are correct and update instantly as user changes info.

**Actual behavior:**
A, G, C: Update as expected whenever you change the quantity or price of a line item.
B, H: Only update if you refresh the browser and then view the invoice again.
D, E, F: Show zero no matter what.

invoice calculations

**OS and Browser:**
Ubuntu 16.04, Chromium, Firefox

I tried to figure this one out but the computed properties are pretty confusing. I would keep on it if somebody wants to help though. Also, the line item "Delete" button doesn't work on this page because it's trying to call an action showDeleteModal() like a function from inside another action, whereas if you replace it with send(), it works.

MatthewDorner:
Fixes #1440.

This PR reverts the change to amountOwedChanged from 4b313ef. It had been changed from an observer to a computed property, but the property is never accessed anywhere so it doesn't get updated correctly.

### DIFF
--- a/app/invoices/edit/controller.js
+++ b/app/invoices/edit/controller.js
@@ -133,7 +133,7 @@ export default AbstractEditController.extend(NumberFormat, PatientSubmodule, Pub
     },
 
     showDeleteItem(itemToDelete, deleteFrom) {
-      this.showDeleteModal(itemToDelete, EmberObject.create({
+      this.send('showDeleteModal', itemToDelete, EmberObject.create({
         confirmAction: 'deleteCharge',
         deleteFrom,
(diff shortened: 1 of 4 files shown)


MatthewDorner:
Let me know if you don't like the "Calculate totals" test, it is my attempt to test the calculated fields logic on the edit invoice page, but not sure it's in line with the style of other tests in the project. Also takes like 10 seconds to run.

PR #1447, opened 2018-06-13
Fix #1429.

### LINKED ISSUE #1429: Adding New Note in history fails
**Expected behavior:** Adding patient history results in the history being visible and accessible through the patient record.

**Actual behavior:** Adding history fails to update the record. New note is not added to the record.

**Steps to reproduce:** Select a current patient, select history tab and +New Note, type in history, click add, refresh page

**OS and Browser:** Linux Mint 18.2,  Firefox 59.0.2

MatthewDorner:
Fixes #1429.

The bug happened because the patient/notes/template.hbs modal dialog allows a user to select which visit to save a note to. The visit needs to be updated when you add/remove a note since it contains a property with a list of associated notes. The bug is that only the visit controller's current (model) visit was being updated, whereas if you select a different visit, you need to update both visits.

The PR adds a conditional check so if you are saving a note with a different visit than the current one on the visit controller, both visits will be updated.


MatthewDorner:
One thing is still a little awkward to me:
- Open Visit A
- Add a note, but change the note's Visit to Visit B
- Save the note
- Visit A will still get updated, even though user didn't specifically ask to update Visit A and they didn't add anything to it. It's possible that if they changed text in other fields (maybe by mistake since they apparently opened the wrong Visit) the Visit would get updated and they would not realize they had done it.

Then again, even if the user doesn't change the selected Visit on a note, there is nothing notifying the user that adding a note has updated the entire Visit record, unlike the message that appears when you click "Update" on the Edit Visit page. So maybe it's a larger question.

Another solution here could be to disable the Visit selection if the note is being added from within a Visit, since the other Visit add-ons don't offer this selection and it appears to be included here only because it's possible to add notes from Edit Patient when no Visit has been loaded.

MatthewDorner:
You know, thinking about it more I think the other solution is better (only show Visit select when adding note from Edit Patient). It doesn't seem like moving notes around between visits was an intended feature. I will close for now and try to code it that way instead.

MatthewDorner:
OK, I changed it. I think it's simpler and more consistent with the way other Visit addons work. If you think the first way is better just let me know.

tangollama:
Thx @MatthewDorner for the PR.

PR #1456, opened 2018-07-17
[Work in Progress]: Use translation file for Invoices and Pricing

### LINKED ISSUE #1437: Invoices and Pricing do not use translations file
Steps to reproduce: review pricing/route.js and invoices/route.js, strings are hard coded. There is a section in the translations file for "billing" but the required strings are missing.

Lily418:
Fixes  #1437 

- Use translationMacro for invoices and pricing

This is a work in progress to allow others to see progress on this issue and give feedback.

### DIFF
--- a/app/pricing/route.js
+++ b/app/pricing/route.js
@@ -1,4 +1,6 @@
 import AbstractModuleRoute from 'hospitalrun/routes/abstract-module-route';
+import { translationMacro as t } from 'ember-i18n';
+
 export default AbstractModuleRoute.extend({
   addCapability: 'add_pricing',
   additionalModels: [{
@@ -16,7 +18,7 @@ export default AbstractModuleRoute.extend({
   }],
   allowSearch: true,
   moduleName: 'pricing',
-  newButtonText: '+ new item',
+  newButtonText: t('buttons.newItem'),
   sectionTitle: 'Pricing',
   subActions: [{
     text: 'All Pricing Items',


pmbanugo:
@Lily418 seems some test are failing. You can run `ember test` to see what's failing.

app/invoices/route.js
```diff
@@ -57,32 +59,32 @@ export default AbstractModuleRoute.extend(ModalHelper, PatientListRoute, {
     }
   },
 
-  subActions: function() {
+  subActions: Ember.computed(function() {
     let actions = [{
-      text: 'Billed',
+      text: this.get('i18n').t('billing.navigation.billed'),
```
stukalin: Just wondering if you can use this `t` alias right away here? What is the difference then in `t('abc')` and `this.get('i18n).t('abc') ?`

Lily418: I'm following the pattern discussed in this issue, I'm not exactly sure why it's required as I haven't used Ember before but intitally I was using t right away and encounted the error described here.
https://github.com/jamesarosen/ember-i18n/issues/352

stukalin: Ok, just good to know, thx

donaldwasserman: Just as an FYI, I think it's mostly to do with `this` and the context of the array inside of an es5 object method vs the newer es6 style `thing: function(){}` vs `thing(){}` + some ember object magic thrown in there for good measure.

PR #1458, opened 2018-07-18
Use fallback translations

### LINKED ISSUE #1457: Translation Synchronization script adds empty strings when no translation is avaliable
**Expected behavior:**
When no translation is avaliable there should be a fallback language used. It should also be easy for contributors to see what requires translation.

**Actual behavior:**
The localisation is filled with an empty string which makes the interface difficult to interact with.

**Steps to reproduce:**
Change language to Russian, observe that many interface items have empty labels.

screen shot 2018-07-18 at 11 45 25

**OS and Browser:**
All

Lily418:
Fixes #1457 

The problem I encounted when creating this PR is that Ember-i18n's [Default Locale Fallback](https://github.com/jamesarosen/ember-i18n/pull/478) works when the key is missing from the translation but I would prefer to include the key and provide a value which allows translation contributors to see more easily what translation keys are defined.

I cannot use empty string as has previously been used because Ember-i18n will not call missing message so instead I've updated this to be undefined. 

Then I added a function for missing message which uses the default locale which I took from here https://github.com/jamesarosen/ember-i18n/issues/256#issuecomment-266221789

- Use undefined rather than empty string when translation is missing, this means result of missingMessage function will be used rather than empty string
- Add ```app/utils/i18n/missing-message.js```
- Update contibution guide to reflect the above change


Lily418:
Tests are currently failing because of this issue with the newer version of ember-i18n in Electron.
https://github.com/jamesarosen/ember-i18n/issues/480

I imagine this is also the reason  #1428 is failing. 

For now I'm going to revert to the older version of ember-i18n as that wasn't the main change here.

PR #1466, opened 2018-07-27
Remove Bower

### LINKED ISSUE #1465: Remove Bower
Bower has been deprecated, is minimally used, which presents a challenge to getting started for new contributors. 

I have a PR that's a work in progress that is moving the following packages need to npm: 

- [x] Bootstrap
- [x] JsBarcode (handled by addon)
- [x] filer.js
- [x] idb.filseystem
- [x] octicons
- [x] pikaday
- [x] typeahead.js
- [x] pouchdb-load
- [x] pretender
- [x] fullcalendar (handled by addon)
- [x] fullcalendar-scheduler (handled by addon)
- [x] timekeeper (used for tests only)
- [x] webtrc-adapber (not used possibly)

donaldwasserman:
Fixes #1465.
See issue for detailed package breakdown. 

Notes: 
* I'll be updating this for the next week or so, as it takes a bit of time to do the migrations, update the dependencies and make sure things don't break
* I've been tempated to use the new auto-import thing, but that seems to be breaking as many things as it helps, we'll see


bower.json
```diff
@@ -1,20 +1,3 @@
...
-    "pouchdb-load": "^1.4.0",
-    "pretender": "~0.10.0",
-    "fullcalendar": "^3.0.0",
-    "fullcalendar-scheduler": "^1.4.0"
-  },
-  "devDependencies": {
-    "timekeeper": "^1.0.0",
-    "webrtc-adapter": "^3.1.3"
-  }
+  "name": "hospitalrun"
```
stukalin: Why do we still need this file? Kill it?

package.json
```diff
@@ -103,15 +104,24 @@
...
     "stylelint-scss": "3.2.0",
+    "timekeeper": "1.0.0",
     "tosource-polyfill": "^0.3.1",
+    "typeahead": "twitter/typeahead.js#v0.11.1",
     "uuid": "^3.0.0",
+    "webrtc-adapter": "^3.1.3",
     "worker-pouch": "2.1.0"
   },
   "dependencies": {
+    "broccoli-funnel": "^2.0.1",
```
stukalin: Why were so many packages moved to `dependencies` and not in the `devDeps`? I'd say we deliver our assets which already contain everything bundled; probably it's not worth to explicitly require some packages to be installed afterwards. Correct me if I'm wrong.

stukalin:
@MatthewDorner probably the same error we had with our `ember-validations`. After `#` commit-ish should go a Git tag, not an npm version or whatever.

package.json
donaldwasserman: TBH, it's because I can never remember the difference in commands between yarn add and the `npm i --save-dev`.

I assume the electron stuff is a requirement there for how the electron builds happen. I moved everything up to dependencies minus the electron stuff.

donaldwasserman:
@stukalin All the tests are covered (with the exception of the barcode one, which requires a configuration setting to be toggled to connect to a printer, which if you have an idea of how to handle that, let me know). 

Just updated master, hopefully things are still green 🤞

MatthewDorner:
Still getting this one when I try to npm install, but if I use yarn it will work.
extraneous

donaldwasserman:
So crashing building leveldown is related to that python issue. You need to
have `python2` available in your path. I'm sure you installed python3.

Not sure what's going on here with npm.

MatthewDorner:
I think it's this, https://stackoverflow.com/questions/43081427/private-github-repo-as-dependency-is-extraneous-on-npm-install, I'm testing to see if I can fix it.

MatthewDorner:
I changed the names in package.json and ember-cli-build to what they are in the package.json of those actual repos (to pikaday-time instead of pikaday and typeahead.js instead of typeahead). I get farther but still get this error unless I use yarn.
barcode

Could just tell people to use yarn?

donaldwasserman:
@MatthewDorner  That's the issue. 

The barcode.js issue you're running into is a problem with an upstream package.

donaldwasserman:
@MatthewDorner @stukalin You guys see any other issues?

MatthewDorner:
nope

donaldwasserman:
Alright, I think it's all good. Just updated, npm installed, yarn installed, tests passed locally, hopefully they'll pass in travis...

Moving forward here, I think.

PR #1490, opened 2018-08-22
Fix issue #1484

### LINKED ISSUE #1484: Appointment can be created with start date later than end date.
**Expected behavior:**
Start date of appointment should be equal or earlier than end date.

**Actual behavior:**
Validation is only performed on end date, so you can enter a valid end date and then change the start date to a later date.

.**Steps to reproduce:**
- Create an appointment of type "Admission"
- Set end date to a date later than start date
- Set start date to a date later than end date
- Save the appointment

appointment

arch-lord:
Fixes #1484.

- Add a new validation script to `startDate` in `app/models/appointment.js`


package.json
```diff
@@ -84,7 +84,7 @@
     "ember-sinon-qunit": "3.2.0",
     "ember-source": "2.18.2",
     "ember-truth-helpers": "2.0.0",
-    "ember-validations": "HospitalRun/ember-validations#2.0.0",
+    "ember-validations": "2.0.0-alpha.5",
```
stukalin: That's a workaround. Shouldn't be there

app/models/appointment.js
```diff
@@ -87,7 +87,36 @@ export default AbstractModel.extend({
...
+              if (endDate.getTime() <= startDate.getTime()) {
+                return true;
+              }
+            }
+          }
+          // patient is properly selected; don't do any further validation
+          return false;
+
+        },
+        message: 'Please select a start date earlier than the end date'
```
stukalin: Can we use i18n here?

package.json
arch-lord: I forgot I put that in there. I'll remove it.

app/models/appointment.js
arch-lord: The other messages in that file don't seem to use it. Should I change all of them?

package.json
arch-lord: Done.

app/models/appointment.js
arch-lord: Absolutely! I'll do both.

MatthewDorner: You can probably do the function the same way as the visit.js model has it with the dateAcceptance function.

MatthewDorner:
@stukalin looks good?

MatthewDorner:
oops looks like one more issue, maybe I can test this in unit/models instead

PR #1495, opened 2018-08-28
Create .python-version

RigoTheDev:
- Create a `.python-version` file with contents: `2.7.15`

When running a fresh `npm install`, it will crash if Python 3 is being used. The reason it's crashing is because `node-gyp rebuild` is attempting to run `python2`, but some systems use Python 3 as their default.

In my case, I was running Python 3.7.0 under [pyenv](https://github.com/pyenv/pyenv).

Here's the error I got:
```
> node-gyp rebuild

gyp ERR! configure error
gyp ERR! stack Error: Command failed: /Users/rigo/.pyenv/shims/python2 -c import platform; print(platform.python_version());
gyp ERR! stack pyenv: python2: command not found
gyp ERR! stack
gyp ERR! stack The `python2' command exists in these Python versions:
gyp ERR! stack   2.7.15
```

Creating a `.python-version` with 2.7.15, would force pyenv to use Python 2 when inside the `hospitalrun-frontend` folder, thus successfully running `node-gyp rebuild`.


.python-version
```diff
@@ -0,0 +1 @@
+2.7.15
```
stukalin: A quick question: wouldn't it force the environment to use exactly `2.7.15`? What would happen if I have, for instance, only `2.7.0` installed?

RigoTheDev: @stukalin Good catch! I'll try to find a solution for that

stukalin: Ah, wait, this is the file to manage the `pyenv`. Weeeellllll, I really don't know if it's really worth it to add a file for a tool which selects a tool to build a tool. I'd personally be against that. You can always use this directive in your working folder and don't commit. Opinions @MatthewDorner @donaldwasserman ?

RigoTheDev: @stukalin That's true, maybe we could add it in the `.gitignore` so that users can set that on their own if needed?

RigoTheDev: This also caused the CI to fail, so I agree it wouldn't be a good idea.

MatthewDorner: so if systems have Python 3 as default, this file alone will not fix it because they still won't have pyenv or the 2.x version of Python? If so then I think I would agree with @stukalin .

RigoTheDev: @MatthewDorner Yep, that seems to be the case here

donaldwasserman:
So I found this issue here, hoping that maybe this would fix itself before 2020 (pytjon2 EOL) which is not the case. 

https://github.com/nodejs/node-gyp/issues/193

One comment included a suggestion to use npm config, I don’t know if that’s a path forward or not. 

What do most computers ship with out of the box? (I do a lot of python, so I really have the curse of knowledge here)

RigoTheDev:
@donaldwasserman I would say Python 2, but a lot of people are migrating to Python 3. Django 2.0 now only supports Python 3 for example

stukalin:
I've got the feeling, that the best solution would be to update our readme file, somewhere in the troubleshooting section and put a couple of solutions there

RigoTheDev:
@stukalin I agree 👍

RigoTheDev:
@stukalin Should I get started on that?

stukalin:
@RigoTheDev you're welcome to start

RigoTheDev:
@stukalin All done 😄

PR #1512, opened 2018-09-17
Add Appveyor CI Configuration

donaldwasserman:
This PR adds appveyor's CI/CD configuration. (Appveyor tests on windows, insdead of just linux). 

This is a first step to automating the deployment of the built electron assets.

Hopefully, this will also have the side benefit of providing more stability for windows users.


MatthewDorner:
I set up a Windows environment and tried to make it the same as the test, but it's passing consistently, unlike here where it's failing. Tried to create my Appveyor account but it keeps giving me an error, can you try restarting the test and see if it fails again?

I've seen this with Travis, where some tests fail intermittently, but it's never been this particular test.

donaldwasserman:
@MatthewDorner Thanks! Worked like a charm.

MatthewDorner:
Still failed, do you have some different settings than me? Mine is pretty much all default.. maybe try clearing whatever appveyor has cached?

Then since we're doing this: https://github.com/HospitalRun/hospitalrun-frontend/pull/1525, should we also change this appveyor script to use yarn? The Travis one already is using yarn, and I guess I should then change the docker script in hospitalrun-server to use yarn and a yarn.lock file as well?

donaldwasserman:
yeah - I'm trying to get it to run the electron specific tests, but I'm not sure why those are specifically failing. 

The bad news is that I looked at the ghost electron app (which uses the same ember-electron setup) and they don't even run the tests....

donaldwasserman:
So just as I said this, it looks like this is a package issue: https://github.com/felixrieseberg/ember-electron/pull/367

donaldwasserman:
I'm closing this and starting over, lol.

Issue #1139, opened 2017-07-27
Missing "Type" field from both Lab Pricing and Image Pricing Screens.


garylm:
**Expected behavior:**  All Pricing Screens under Billing should have a place to set the "Type".

**Actual behavior:**  Lab and Imaging do Not have the Type field appearing on the Add or Edit functions, but the lists have a column for Type.  In comparison both Procedure and Ward Pricing do have the Type field on their Add/Edit screens.

Here is a screen shot of Imaging Pricing screen w/o Type
[OWH Image Pricing Screen wo Type.pdf](https://github.com/HospitalRun/hospitalrun-frontend/files/1180600/OWH.Image.Pricing.Screen.wo.Type.pdf)

Here is the Imaging Pricing List showing Type as a Column
[OWH Image Pricing List showing Type column.pdf](https://github.com/HospitalRun/hospitalrun-frontend/files/1180602/OWH.Image.Pricing.List.showing.Type.column.pdf)

Here is a screen shot of Procedure Pricing screen with Type
[OWH Procedure Pricing Screen w Type.pdf](https://github.com/HospitalRun/hospitalrun-frontend/files/1180601/OWH.Procedure.Pricing.Screen.w.Type.pdf)

**OS and Browser:**  Chrome and Firefox, running Docker Server on Ubuntu

KenyanGeek:
@garylm experiencing same issue. 'Type" used to be there but seems to have disappeared and hence not possible to add "billable" Lab or Imaging charges on the invoice. @jkleinsc any reason where this was deprecated?

tangollama:
@garylm I'm looking into this now.

tangollama:
@garylm @KenyanGeek this was an intentional decision in the system to limit the type field for Lab and Imaging prices (see: https://github.com/HospitalRun/hospitalrun-frontend/blob/master/app/pricing/edit/controller.js line 102). Can you articulate a reason why we should make this change? If so, it's a pretty easy change and PR.

KenyanGeek:
Hi @tangollama thanks for your response. In the case of a Hospital with an extensive list of procedures it would be useful to be able to pre-input these procedures from a csv. Because otherwise the end user is confused on how this works and how it connects to the billing at the end. I hope that makes sense?

tangollama:
@KenyanGeek I'm confused by your response. What do lists of Procedures have to do with a type field being present on Imaging and Lab Prices?

KenyanGeek:
[...]

Issue #877, opened 2016-12-15
Image and Lab Pricing Types - default "Procedure" value should not be editable 


mnorbeck:
On the Imaging Pricing and Lab Pricing Type Pages, the default values of "Lab Procedure" and "Image Procedure" should not be editable.   Editing them causes problems

PR #1526, opened 2018-10-11
Pricing types

### LINKED ISSUE #1444: Cannot add Imaging Pricing Types manually
**Expected behavior:** Edit Lookup Lists, select Imaging Pricing types, add pricing type

**Actual behavior:** No option to add a pricing type is available, only file upload.

**Steps to reproduce:** Select Lookup lists from Admin menu, go to Imaging Pricing Types, try to add new pricing type

screenshot from 2018-05-30 01-21-21

**OS and Browser:** Hospitalrun-server (Ubuntu 16.04) accessed via Linux Mint 18.3 and Firefox 60.0.1

MatthewDorner:
Fixes #1444

https://github.com/HospitalRun/hospitalrun-frontend/issues/877
https://github.com/HospitalRun/hospitalrun-frontend/commit/aa09e64951859592804ff541e31dda5b2b058ada
https://github.com/HospitalRun/hospitalrun-frontend/issues/1139

Based on the previous changes and discussion above, the intention was to remove user's ability to set Pricing Type for Pricing models with Category == 'Lab' or Category == 'Imaging' and instead lock them to default value of 'Lab Procedure' or 'Imaging Procedure'. But, the lookup lists for these weren't fully removed which has confused users and the pricing/edit controller had a bug which meant the Pricing Type would sometimes be set to blank instead of the appropriate default value, and this caused that Pricing model to become unusable due to the Lab and Imaging Request pages filtering out anything that didn't match the default 'Lab Procedure' or 'Imaging Procedure'.

In this PR, I removed the lookup lists and fixed the pricing/edit controller behavior. I also removed unused code related to the old behavior (with the lookup lists). Also removed the labs/charge and imaging/charge routes entirely as they're not used now and I believe they'd need to be re-written if they were to be used. I think they were intended for adding lab or imaging charges from another route via a modal dialog.

Also changed pricing/edit controller so when you have Category set to 'Lab' or 'Imaging', it shows Pricing Type as a static text field set to the appropriate default value instead of  just making it disappear as it did before.

I will add some test cases in the next couple days.


MatthewDorner:
One other thing, you can see that I removed the reference to 'imaging_pricing_types' from the imaging/route.js' additionalModels array. But I couldn't do the same thing for labs/route.js. That's because in labs/route.js, there was only that one element in additionalModels and I found that removing it causes a problem.

Basically, querying for the additionalModels in abstract-module-route's model(), even if it returns nothing, has some effect of initializing relational-pouch and if you don't do it, the code in abstract-index-module's model() will fail. I will create a new issue for that though.

PR #1530, opened 2018-10-12
Sign in with Google should be an additional button.

### LINKED ISSUE #1529: UseGoogleAuth takes over all authentication options
**Expected behavior:**
As a user, when a site supports a 'Sign in with Google' option it's usually in addition to using a username and password and I expect to be able to click the 'Sign in with Google' button. 

**Actual behavior:**
When useGoogleAuth is enabled the login screen automatically redirects to the Google OAuth flow.

This also limits our ability to add another authentication provider like https://www.npmjs.com/package/passport-doximity

ATLCTO:
Fixes #1529.

Don't force a redirect to '/auth/google' until the user clicks a 'Sign in with Google' button.

### DIFF
--- a/app/controllers/login.js
+++ b/app/controllers/login.js
@@ -7,6 +7,7 @@ let LoginController = Controller.extend({
   errorMessage: null,
   identification: null,
   password: null,
+  useGoogleAuth: null,
 
   actions: {
     authenticate() {
@@ -23,6 +24,10 @@ let LoginController = Controller.extend({
           this.set('offlineError', false);
         }
       });
+    },
+
+    authenticateGoogle() {
+      window.location.replace('/auth/google');
     }
   }
 });
--- a/app/locales/en/translations.js
+++ b/app/locales/en/translations.js
@@ -940,7 +940,8 @@ export default {
     labels: {
       password: 'Password',
       signIn: 'Sign in',
-      username: 'Username'
+      username: 'Username',
+      signInGoogle: 'Sign in with Google'
     },
     messages: {
       error: 'Username or password is incorrect.',
--- a/app/routes/login.js
+++ b/app/routes/login.js
@@ -25,9 +25,7 @@ export default Route.extend(UnauthenticatedRouteMixin, {
         });
       } else {
         return get(this, 'config').useGoogleAuth().then((useGoogleAuth) => {
-          if (useGoogleAuth) {
-            window.location.replace('/auth/google');
-          }
+          this.controllerFor('login').set('useGoogleAuth', useGoogleAuth);
         });
       }
     }
--- a/app/styles/views/_sign_in_screen.scss
+++ b/app/styles/views/_sign_in_screen.scss
@@ -33,4 +33,23 @@
   font-weight: 300;
 }
 
-.form-signin-heading::first-letter { text-transform: uppercase; }
+.btn-google-sign-in {
+  margin-top: 15px;
+  background: #4285f4;
+  width: 100%;
+  color: $white;
+
+  &:hover,
+  &:focus {
+    background-color: #2268f1;
+  }
+
+  &:active,
+  &.active {
+    background-color: #5491f5;
+  }
+}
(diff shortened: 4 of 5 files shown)


ATLCTO:
image

MatthewDorner:
The code change is pretty simple, I'm trying to think of any implications. When I see "Sign in with Google" usually I take it to mean a different type of account that can only sign in with Google. But here  the account is always created by the HR admin in 'New User', and if you use this setting (with the changes in this PR) you can still sign in with google, OR with the password created on the HR 'New User' screen, for any given account. Is that right?

I notice that if you don't enter a password in 'New User', it creates a random one, and this apparently is what you would do for users who are signing in only via Google, but it's not made obvious that you should. We could consider a modification made to New User that allows you to set the means by which the user is allowed to login.

Just my thought, admittedly I don't understand authentication systems much so I will refrain from approving/denying.

ATLCTO:
Long term I was thinking we could add additional authentication providers, so it would look more like other sites-

image

[Federated identity](https://en.wikipedia.org/wiki/Federated_identity) could be integrated so a hospital can grant all their users access by connecting the two system together and not have to manually create each user account, etc. 

I already implemented https://github.com/CKWC/passport-doximity so that it could also be implemented in the future. I was thinking you all would prefer, smaller pull requests than one massive pull request with a bunch of things added.

MatthewDorner:
@donaldwasserman any input on this?

donaldwasserman:
I have been giving this some thought.

I can't decide if the current behavior is a feature or a bug. 

The way I've always thought the "login with google" functionality _should_ work is similar to that of Slack's login with domain feature, aka: Single Sign On, just outsourced to google. For example: I only want users at `<name>@mydomain.com` to be able to sign on, and I use G Suite for my domain. 

The Other option, and they way @ATLCTO see's it, is like the way tons of different services work, where you are allowed to login with service X. 

Now, looking at **check's notes** the thorough Hospital Run docs, I don't really see what was the _originally intended behavior_. 

There's certainly no code/docs that prefer _my_ interpretation: https://github.com/HospitalRun/hospitalrun-server-routes/blob/master/routes/30-auth.js

This brings up a bigger question of "how do we deal with users" more generally, especially from managing them. 

@ATLCTO What's your take on all of this? Maybe we could have non-github discussion of this as well.

ATLCTO:
Yes, feel free to hit me up on slack. My own motivation was just to get this merged in as easy as possible so I could add the 2nd 'Login with Doximity' option so I could use that (we have 150+ doctors who login with Doximity) but wanted to add the two pull requests separately (but I can combine them). 

I've been reading more on this and it sounds like it's delegated vs federated login- [What is the difference between federated ID, SSO, and delegated ID](https://www.quora.com/What-is-the-difference-between-federated-ID-SSO-and-delegated-ID-with-examples-that-explain-the-concept-clearly/answer/Jitendra-Niberiya). So I guess the real question is delegated vs federated (do we allow just Google for delegated, or multiple providers). It seems like this would need to be configurable, because we shouldn't decide which providers our end-user should trust, so they should be able to manage them... if we allow federated login as a feature. My motivation for Doximity is described in https://www.doximity.com/developers/documentation#oauth

_The Doximity API allows partners to utilize Doximity’s clinician verification engine to verify the identity of a user and allow the user to authenticate using her Doximity credentials. That’s right, one less username and password for a doctor to remember, less friction during registration and confirmation that the user is a clinician._

Don't feel obligated to merge this in, I've already forked all the projects and I'm merging in whatever changes in there, so I'm not blocked.

donaldwasserman:
@ATLCTO Yeah, I think going down the configuration option path is great. There's a lot of good use cases for doing it this way. I've created an RFC section in the planning repo, so maybe this can be our first RFC. I think some thought on how/why/potential changes would be a good idea, but this is definitely a first step. 

I'm just thinking of how this would work in a low-connectivity solution as well, and provide for some answer to that question.

PR #1545, opened 2018-10-24
Mjprude/nvmrc

### LINKED ISSUE #1488: Get To Node 8/10x, add .nvmrc
This application should, at a minimum work across one of the default Ember supported LTS node releases: 

Node 6.x, 8.x, 10.x

There are a few different addons who are misbehaving and need to be updated in order to meet this use case. 

Also, add a `.nvmrc` file to help automatically use the right node version. [See NVM](https://github.com/creationix/nvm)

mjprude:
Fixes #1488

- Add an .nvmrc to point nvm users to a supported node version
- Update README point to 6.14.3+
- Update yarn.lock - in my attempts to get `yarn install` to run, I upgraded to latest yarn (`v1.10.1`) which altered the yarn.lock file. Happy to exclude this commit since it is not directly related to this PR.


MatthewDorner:
Sorry for not getting to this earlier. Node 6.x will be out of support in a few months https://nodejs.org/en/about/releases/, it would be better to get to a newer version.

After updating ember-cli-babel which explicitly won't work with 6.x, I'm still seeing other errors on both latest 8.x and 10.x.

MatthewDorner:
Specifically after updating ember-cli-babel on latest 8.x, I get "Invalid Option: optional is not a valid top-level option." on `npm start`. This is referring to the `optional: ['es6.spec.symbols']` line in ember-cli-build.js, I think it is related to this: https://github.com/ember-decorators/ember-decorators/issues/90. 

It looks like we need to use this: https://babeljs.io/docs/en/babel-plugin-transform-typeof-symbol instead of es6.spec.symbols and add it in the way described in the previous github link.

Would you like to continue working on this in this PR?

PR #1551, opened 2018-10-31
Fix 1543 report end date

### LINKED ISSUE #1543: incident report doesn't display current date if no end date is selected
When generating an incident report without an end date, the report does not display the current date for the end date in the report title or the file name of the exported report. I think this is especially confusing if a user saves this file for records, since it will not have an end date.  

**Expected behavior:**
Unsure, however I think that either of the following would make sense: 
1. Require both start and end date and not allow a report to be generated if one is not supplied and display an error to the user
2. Display the current date in the report title and exported report

I think two makes the most sense because it seems that it was deliberately coded to make it so an end date would not have to be selected. 

**Actual behavior:**
End date is not displaying in the title if none is selected.

**Steps to reproduce:**
1. Go to Incidents -> Reports
2. Select a start date
3. Do not select an end date
4. Generate Report
5. Look at report title
6. Export Report
7. Look at report file name

image

**OS and Browser:**
MacOs, Chrome

jackcmeyer:
Fixes #1543 

* adds functionality for reporting modules to automatically select an end date if one was not chosen. It will default to the current date based on the user's system. 
* updates relevant tests

This autofill functionality is consistent with the Inventory Valuation and Inventory by Location Reports where the single date is auto filled to the current date based on the user's system if one was not provided. 

### DIFF
--- a/app/incident/reports/controller.js
+++ b/app/incident/reports/controller.js
@@ -63,21 +63,23 @@ export default AbstractReportController.extend(UserSession, NumberFormat, {
   _findIncidentsByDate() {
     let filterEndDate = get(this, 'endDate');
     let filterStartDate = get(this, 'startDate');
+
     let findParams = {
       options: {},
       mapReduce: 'incident_by_date'
     };
-    let maxValue = get(this, 'maxValue');
-    return new RSVP.Promise(function(resolve, reject) {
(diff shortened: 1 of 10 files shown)


jackcmeyer:
I think I completely hosed the git history on this PR.. I'm going to close it and open a different one

PR #1552, opened 2018-10-31
Fix 1543

### LINKED ISSUE #1543: incident report doesn't display current date if no end date is selected

jackcmeyer:
Fixes #1543 

* adds functionality for reporting modules to automatically select an end date if one was not chosen. It will default to the current date based on the user's system. 
* updates relevant tests

This autofill functionality is consistent with the Inventory Valuation and Inventory by Location Reports where the single date is auto filled to the current date based on the user's system if one was not provided. 

### DIFF
--- a/app/incident/reports/controller.js
+++ b/app/incident/reports/controller.js
@@ -63,21 +63,23 @@ export default AbstractReportController.extend(UserSession, NumberFormat, {
   _findIncidentsByDate() {
     let filterEndDate = get(this, 'endDate');
     let filterStartDate = get(this, 'startDate');
+
     let findParams = {
       options: {},
       mapReduce: 'incident_by_date'
     };
-    let maxValue = get(this, 'maxValue');
-    return new RSVP.Promise(function(resolve, reject) {
-      if (isEmpty(filterStartDate)) {
-        reject('Start date cannot be an empty value.');
-      }
-      findParams.options.startkey =  [filterStartDate.getTime(), null];
 
-      if (!isEmpty(filterEndDate)) {
-        filterEndDate = moment(filterEndDate).endOf('day').toDate();
-        findParams.options.endkey =  [filterEndDate.getTime(), maxValue];
+    let maxValue = get(this, 'maxValue');
+
+    return new RSVP.Promise(function(resolve, reject) {
+      if (isEmpty(filterStartDate) || isEmpty(filterEndDate)) {
+        reject('Start or End date cannot be an empty value.');
       }
+
+      findParams.options.startkey =  [filterStartDate.getTime(), null];
+      filterEndDate = moment(filterEndDate).endOf('day').toDate();
+      findParams.options.endkey =  [filterEndDate.getTime(), maxValue];
+
       return this.store.query('incident', findParams).then(resolve, reject);
     }.bind(this));
   },
@@ -145,6 +147,12 @@ export default AbstractReportController.extend(UserSession, NumberFormat, {
 
   actions: {
     generateReport() {
+
+      if (isEmpty(get(this, 'endDate'))) {
+        let now = new Date();
+        this.set('endDate', now);
+      }
+
       let reportRows = get(this, 'reportRows');
       let reportType = get(this, 'reportType');
       reportRows.clear();
--- a/app/patients/reports/controller.js
+++ b/app/patients/reports/controller.js
@@ -855,6 +855,11 @@ export default AbstractReportController.extend(PatientDiagnosis, PatientVisits, 
 
   actions: {
     generateReport() {
+      if (isEmpty(this.get('endDate'))) {
+        let now = new Date();
+        this.set('endDate', now);
+      }
+
       if (this._validateDates()) {
         let reportRows = this.get('reportRows');
         let reportType = this.get('reportType');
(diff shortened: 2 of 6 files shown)


app/incident/reports/controller.js
```diff
@@ -145,6 +147,12 @@ export default AbstractReportController.extend(UserSession, NumberFormat, {
 
   actions: {
     generateReport() {
+
+      if (isEmpty(get(this, 'endDate'))) {
+        let now = new Date();
```
donaldwasserman: Should this me a moment object or raw Date?

cc: @MatthewDorner whats your thought?

app/patients/reports/controller.js
```diff
@@ -855,6 +855,11 @@ export default AbstractReportController.extend(PatientDiagnosis, PatientVisits,
 
   actions: {
     generateReport() {
+      if (isEmpty(this.get('endDate'))) {
+        let now = new Date();
```
donaldwasserman: See above

app/incident/reports/controller.js
MatthewDorner: I believe you'd have to turn it into a Date anyway as that's what the Ember date attribute uses, so it seems ok. I'm more wondering if we need to think about time zones.

MatthewDorner: Note that we also changed the behavior of the DB query here, as it used to leave the end date open meaning it would get everything but now we set the end date to the end of current day as it is in user's local time (I think).

MatthewDorner:
I did some testing and I think this is ok. Local time works the same here as it does everywhere: the datetimes the user sees/inputs are in user's local time, but those stored in database are converted to universal. And in the particular case of report queries, the start date of the query already behaves this way so I think it's fine if end date also does.

It does mean users can get different results for queries that appear the same if they're in different time-zones, but if there's a "solution" to that it should be in documentation or in a separate issue to display time zones or something.

jackcmeyer:
@MatthewDorner I can make these changes. It appears that there haven't been error messages on the incident and inventory screens previously and the autopopulate end date bug is something that I overlooked. 

Couple of questions:

1. Do we want modal error messages?
2. What should the error message read?

MatthewDorner:
I feel like the way it is in patient is fine, the same modal dialog with the same message. The only thing that should be checked for is start date, so I guess you can remove the check for end date there as well right?

I see "Inventory By Location" and "Inventory Valuation" don't use a Start Date, just an End Date (which is labeled as Effective Date) so we should disable the check for that one the same way Patient disables it for "Patient Status" report.

Thanks.

app/patients/reports/controller.js
```diff
@@ -838,14 +838,28 @@ export default AbstractReportController.extend(PatientDiagnosis, PatientVisits,
     let reportType = this.get('reportType');
     let startDate = this.get('startDate');
     if (reportType === 'status') {
+      if (isEmpty(endDate)) {
+        let now = new Date();
+        this.set('endDate', now);
+        endDate = this.get('endDate');
+      }
```
MatthewDorner: Is this added code necessary?

jackcmeyer: No, I don't think so. I was mistaken thinking status reports needed an end date. I've fixed it.

MatthewDorner: thanks, I noticed something else too.

In patient, try selecting Admissions Detail, then set both dates to some date range, then go back to Patient Status and generate. As long as the properties are already set to something, it will still show a date range in the report title. Or if only one of the dates is set, that one will appear. Really, I am thinking we want no dates to ever appear in Patient Status title.

The two Inventory reports that don't have start date, don't seem to have this problem. You can set both dates on one report, then switch to Inventory Valuation but it will still only show the end Date.

jackcmeyer: Looks like this is an existing issue! I'll take a stab at fixing it. 

Do you think just setting the start and end date to null when the status report is generated is sufficient?

app/patients/reports/controller.js
```diff
@@ -276,6 +276,8 @@ export default AbstractReportController.extend(PatientDiagnosis, PatientVisits,
 
   isStatusReport: computed('reportType', function() {
     let reportType = this.get('reportType');
+    this.set('startDate', null);
+    this.set('endDate', null);
```
MatthewDorner: With this, it erases the dates every time you change the Report Type. Although it's funny, they still look like they're filled in in the text inputs. But the actual properties are erased, as if you set the dates on Admissions Detail, switch to Admissions Summary and then click Generate Report and it'll say your start date is missing.

You could just make it conditional so it only sets these if reportType === 'status'.

jackcmeyer: the latest commit should fix that issue.

MatthewDorner: cool, thanks for making all those revisions but I think we fixed like 3 issues here, so good job!

PR #1553, opened 2018-11-04
Migrate to ember-intl from ember-i18n

### LINKED ISSUE #1524: Replace Ember I18n with Ember-Intl
The Internalization system in Hospital Run works quite well right now for strings. However, in the time since HR originally was written, there have been several developments in both browser API's and technologies that changed some of the underlying assumptions of Ember-Intl. 

There are several points of difference between the Ember-I18N and Ember-Intl: 

- Ember I18n doesn't support ICU syntax and isn't aware of new browser locale API's
- Ember I18n [plans to deprecate itself](https://github.com/jamesarosen/ember-i18n/issues/481)
- Ember-Intl also handles date, relative time, and number formatting
- Ember-Intl also supports asynchronous loading of translation files

There are several outstanding issues that could be solved by this: #990, #1501, #1430 etc 

This is somewhat of a major effort, but there may be a [community driven codemod solution outlined here](https://github.com/jamesarosen/ember-i18n/issues/481)

chadian:
Fixes #1524.

- Migrates translations to `json` format used by `ember-intl`
- Fix usages of translation in application of `ember-intl`
- Fix tests to be setup with `ember-intl` and its translations
- Remove `ember-i18n` and its existing translations

Note: Some of the translations are an incomplete set, and might need additional cleanup. `ember-intl` will actually report these missing translations across the files at build time.


app/macro.js
```diff
@@ -0,0 +1,6 @@
+import { writable } from 'ember-macro-helpers';
+import { translationMacro } from 'ember-intl';
+
+export const t = function() {
+  return writable(translationMacro(...arguments));
```
chadian: Because the computed property that was created with `i18n`'s `t` there was existing logic that allowed calling `.set()` and replacing the existing computed property. However, in `ember-intl` this is set to a read-only property. This wrapper ensures that the `translationMacro` is wrapped and is `writable`.

donaldwasserman:
@chadian - Thanks a bunch this looks great. 

Just going through it first, but looks solid thus far. 

Trying to figure out why the old version of travis is still running.

PR #1556, opened 2018-11-08
Swap ember-browserify for ember-auto-import

### LINKED ISSUE #1544: Swap ember-browserify for ember-auto-import for npm-installed modules
**Expected behavior:**
Use npm modules without `ember-browserify` as it is deprecated and the community is moving to `ember-auto-import`

**Actual behavior:**
Use `ember-auto-import` and drop `npm:` prefixes in imports

**Steps to reproduce:**
N/A

N/A

**OS and Browser:**
N/A

chadian:
Fixes #1544.

`ember-browserify` is being deprecated in favour of `ember-auto-import` which is using webpack behind the scenes. This PR takes care of integrating `ember-auto-import` and will allow anyone to more easily use npm packages directly in the future and potentially avoid the usage of ember shims.

In one specific case with `ember-pouch` we actually want to let it shim `pouch` with a few needed addons and not have the `import` pull the `pouch` package directly but instead from `ember-pouch`.

- Removes `ember-browserify` in favour of `ember-auto-import`
- Uses `pbkdf2` package directly instead of relying on a `crypto-browserify` which was previously included via `browserify`.


donaldwasserman:
@chadian this looks great! @MatthewDorner Any additional thoughts?

MatthewDorner:
nope, looks good.

donaldwasserman:
Actually, I’m going to merge the ember intl one first

MatthewDorner:
I can't login to the application after these changes, using Chromium or Chrome v70 and Ubuntu 18.04. But Firefox v63 is OK. The login page normally refreshes once when you load it the very first time (with your browser data cleared), and then you can log in. In this case, that refresh doesn't happen and the login button doesn't do anything. If you look on the dev tools "Application" tab, and "Service Workers", you see the service worker status is stuck on "installing".

Tried it before and after the update from master I just did, same results.

This has been acting weird for me before. After clearing browser data I usually had to open up Chromium to login screen, close it, then open it up again, then it'll refresh and I can login. And when I recently tested a production build, once I got it running on hospitalrun-server I didn't have to do this at all (but that wasn't with the changes in this PR). I haven't tried making a production build with this PR, but I might do that next. Anybody also seeing this or have any ideas?

MatthewDorner:
Same thing happens for me with production build and hospitalrun-server. Can't log in with Chromium or Firefox. Sorry @chadian it's nothing obviously related to your PR, but it seems like it's causing some change. Unless I'm the only one seeing this. I'll try to figure out what's actually causing it.

chadian:
Hi @MatthewDorner. I can look into this. From the sounds of it, there's definitely a side effect being introduced by this PR and the introduction of `ember-auto-import`? What is the significance of the window refreshing after the service worker installed? I will see what I can find!

MatthewDorner:
I looked into it the past few days. Somehow it's getting stuck on the "install" step of the service worker lifecycle, during which the precache files are loaded. But, I looked at the files being precached and it hasn't changed with this PR, it's the same files. But also, if I hack it so it doesn't precache any files, then it works.

The significance of the refresh is, I think you need to refresh the page once the service worker is ready, in order for the service worker to take control of the page. So since service worker is never ready, it doesn't refresh and app doesn't work.

Also I found that it works fine if I have dev tools open, so it's pretty weird considering that it also doesn't happen on Firefox. The service worker file gets generated in `/dist` folder based on the files in our `app/serviceworkers` folder, plus the `ENV.serviceworker` settings in `config/environment.js`, and it's generated by broccoli-serviceworker which uses sw-toolbox.

If we can't figure it out, we can probably still merge this, since I'm pretty sure the fix will be to fix the original problem which existed before rather than modifying your PR somehow. I just wanted to research since it is now worse than before.

MatthewDorner:
I'd also like to know first if you can reproduce the problem.

chadian:
I wasn't able to reproduce with Chrome 73 in incognito. I attached a GIF showing a fresh incognito window and I was able to get the double refresh on this branch (but had mentioned it also exists on master, too).

doublereload

Maybe it's something related to that specific version of Chrome or Linux? Do you think we can merge this and investigate on a separate issue?

Also I re-ran the tests locally and they should be passing. I think we had a false negative, can we kick off travis ci again?

MatthewDorner:
I agree, since it behaves differently when dev tools are open I am starting to suspect something in the browser. And Travis has been intermittently failing on that particular test, so you're right, it's not related.

PR #457, opened 2016-05-12
Fix issue #380: cascading delete for patient records; delete related records first and then patient record

aboma:
Fixes #380 Cascading Delete for patient records.

**Changes proposed in this pull request:**
- Delete visits, invoices, and most associated records when deleting patient. Delete these records before deleting patient record to avoid missing records errors.
- Add view to return invoices for patient.

cc @HospitalRun/core-maintainers

Issue #381, opened 2016-03-19
Flag patients as archived rather than deleting them entirely


tangollama:
This will involve not only a new status on patients but also adjustments to the various views to exclude patients that are flagged as "archived."

pete-the-pete:
@tangollama i can take a look at this and #382

jkleinsc:
@pete-the-pete I started working on this and I checked in a branch here with my work: https://github.com/HospitalRun/hospitalrun-frontend/tree/archive-records

If you look at the code here, basically what I am doing is adding a new attribute called archived.  Then in the pouch/couch views, any records with that attribute are ignored.  Tests are not passing right now on this branch, but the basic concepts are implemented.  If you want to continue work on it, feel free, but I would use that branch as a starting point.

jkleinsc:
This issue was resolved in commit 20dd5b29cf697b23f0df179d30fe951e31744d0f.

PR #1559, opened 2018-11-15
Cascading deletion for visit

### LINKED ISSUE #1439: Deleting visit breaks any associated invoice to that visit
**Expected behavior:**
Not sure.

**Actual behavior:**
Similar to #383. If you add an invoice, you must associate a visit to the invoice. If you then delete that visit, any listing at /invoices that would have included your invoice just won't render. You will get this error:
invoice visit relationship error

**OS and Browser:**
Ubuntu 16.04, Chromium

MatthewDorner:
Fixes #1439

- Adapted code from https://github.com/HospitalRun/hospitalrun-frontend/pull/457 to also work in visits/delete instead of only in patients/delete.
- Moved code used by both patients/delete and visits/delete into a cascading-deletion.js mixin.
- Rewrote invoice deletion code (deleteInvoicesTask) as it had a bug.
- Updated patient.txt test fixture to include new invoices_by_visit view.
- Fixed so patient and visit records are archived like other records, instead of simply deleted from database, see https://github.com/HospitalRun/hospitalrun-frontend/issues/381.

I am guessing I have merge conflicts related to the translation package switch, and may add better testing, so I'll have a few more updates.


MatthewDorner:
I'm going to try to add a Unit test for the cascading-deletion.js mixin.

MatthewDorner:
Added a unit test, I probably messed up the ember-concurrency stuff though so going to look into that now.

MatthewDorner:
I think this is finally ready. I added some extra relations for deletion:

visit: added 'reports' and 'diagnoses'
patient: added 'allergies', 'diagnoses', 'operationReports', 'operativePlans'
operation-report: added 'diagnoses' and 'preOpDiagnoses'
operative-plan: added 'diagnoses'

MatthewDorner:
Also removed the hasMany('visit') relation on the appointment model. I couldn't find any purpose it served and I looked at the code when it was added, which was as part of a "missed appointments" feature that's now removed (which also didn't appear to even use this relation, though maybe anticipated using it).

PR #1560, opened 2018-11-16
Test suite upgrade

chadian:
- [RFC#232](https://github.com/emberjs/rfcs/blob/master/text/0232-simplify-qunit-testing-api.md)
- [RFC#268](https://github.com/emberjs/rfcs/blob/master/text/0268-acceptance-testing-refactor.md)
- Install `ember-qunit` and Remove `ember-cli-qunit`
- Use `@ember/test-helpers`

I had to implement a `jqueryLength` and `jquerySelect` to use old selectors that aren't compatible with `querySelector`. I would recommend that these be treated as deprecated and in the future we use [`ember-test-selectors`](https://github.com/simplabs/ember-test-selectors) and use `[data-test-*]` selectors and refactor away from selectors that contain `:first()`, `:last()`, `:contains()`, and any other jquery-specific selectors.

As a migration plan we can look at usages of `jqueryLength`, `jquerySelect` and `$`, and refactor those selectors to not be dependent. Also continuing to use `qunit-dom` will help out a lot.

I also noticed that many of the tests needed extra wait helpers to pass. I'm not sure why there was extra lag, it might be due to something not being captured by a run loop.

I recommend we move `runWithPouchDump` into a helper that uses qunit's hooks. This type of functionality will fit more naturally and in bootstrapping the state of the tests. Now with qunit you can nest `module`'s so we can do separate bootstrapping with these hooks, so we might be able to avoid cases where test wrappers are used by passing `assert` and other setup helper functions.

Let me know if there's any questions. Since this touches all the tests I hope we can merge it soon so that it doesn't fall behind.


chadian:
Can someone re-run the tests on travis-ci for this PR?

I triggered a rebuild with an empty commit and [everything seems to be passing](https://travis-ci.com/HospitalRun/hospitalrun-frontend/pull_requests), but it would be nice not to have a garbage commit in the history. Thanks!

After that if I get green lights I think the PR is ready for review.

MatthewDorner:
Sorry it wasn't letting me re-run the tests for some reason but now it is, and I'll start reviewing now too. Can you fix the merge conficts?

MatthewDorner:
ok I think that'll fix it

chadian:
Thanks @MatthewDorner! Looks like you fixed the merge conflicts, let me know if there's anything else. Cheers.

MatthewDorner:
Thanks for the PR @chadian !

Issue #1247, opened 2017-10-04
Right-to-left languages don't trigger a layout change to right-to-left


BogdanAlexandru:
**Expected behavior:**
Right-to-left languages like Urdu change the display layout of text.

**Actual behavior:**
The text is still rendered as if a left-to-right language is used.

**Steps to reproduce:**
Start up HospitalRun.
If Urdu is not selected, select Urdu from the languages dropdown.

**Screenshots (if applicable):**
![Imgur](https://i.imgur.com/5XE2kCJ.png)

**OS and Browser:**
Windows 10 + Chrome.

The layout should be similar to:
![Imgur](https://i.imgur.com/1bRpnto.png)

tangollama:
@BogdanAlexandru that's a big, tall order. Setting this to a help wanted tag.

LinKassem:
I will work on it.

disposedtrolley:
I don't have enough experience with Ember to implement this, but we could possibly modify the `selectLanguage()` method in `language-select.js` to add a `direction: rtl` CSS rule to `div`s. Would also be good to have a mapping of which languages read right-to-left somewhere so we don't end up with a big switch statement.

<img width="1249" alt="screen shot 2017-11-01 at 3 37 24 pm" src="https://user-images.githubusercontent.com/4476568/32260231-a1cbdcca-bf1a-11e7-915e-064cc7165c60.png">

This seems to shift everything properly except for the table header, which has a `text-align: left` rule applied. There's also some tidying to be done re: the spacing between the text and icon in the action buttons.

BogdanAlexandru:
Marking each rtl language as rtl could likely be done in `/<locale>/config.js` as can be seen [here](https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Translation-Compiler).

There's in fact support for the rtl flag in ember-i18n. This is what it does by default: https://en.wikipedia.org/wiki/Right-to-left_mark

MhdTlb:
Can I work on this?

morsh:
Is this one still open for grabs?

tangollama:
Yes. 

Typed or dictated imperfectly on my iPhone (717.385.9970)

morsh:
@tangollama - added Hebrew translation (#1389), added 'rtl' attribute to urdu (#1388) and added support for RTL layout/style (#1388).

PR #1562, opened 2018-11-17
ember-intl implementation fixes

### LINKED ISSUE #1561: ember-intl upgrade side effects
**Expected behavior:**
Warning message should have no HTML in it.
**Actual behavior:**
Warning message has HTML in it.

**Steps to reproduce:**
_New Inventory Item Add_
1. Login
2. Click on Inventory Received 
3. Click on add button
4. View the message with HTML in it

OR 
_New Patient Quick Add_
1. Login
2. Click on Medication -> New Request
3. Enter patient name that doesn't exist
4. Fill in other information
5. View error message

_New Inventory Item Add_
image

_New Patient Quick Add_
image

**OS and Browser:**
MacOS, Chrome 70, Firefox, Safari

- [x] warning messages which use HTML to render are showing the raw html in the message
- [x] unable to select first lanaguage
- [x] navigation bar items not being translated
- [ ] warnings with Hebrew in the console and some Hebrew messages not rendering
- [x] default to English for missing translations
- [x] different results between missing translations and empty translations
- [x] loss of support of right-to-left languages

chadian:
Partially fixes #1561.

* remove empty string translations
* setting a missing translation message to an empty string
* Adding the fallback locale


app/services/language-preference.js
```diff
@@ -1,10 +1,8 @@
 import { run } from '@ember/runloop';
-import { getOwner } from '@ember/application';
 import { hash } from 'rsvp';
 import Service, { inject as service } from '@ember/service';
 import Ember from 'ember';
 import config from '../config/environment';
-import { walkConfigs, setRTL } from '../utils/locale-utils';
```
chadian: These were removed along with the corresponding files since they don't relate to `ember-intl` anymore and were caught in my find/replace for `ember-i18n` to `ember-intl`. Is there a way I can verify that RTL is working properly with `ember-intl` in the same way it was before with `ember-i18n`?

MatthewDorner: I think there was one included locale that used that, Urdu. I remember it caused the entire nav bar to move to the right side of the screen instead of left. There's this https://github.com/HospitalRun/hospitalrun-frontend/issues/1247.

MatthewDorner:
Took a look at the RTL with Urdu language, looks good except I noticed something else: the font changes. It even sometimes changes in the middle of a translation item, and there's no HTML or anything in the translation string, I wonder what it is. And I don't see this in any other locale even other RTL ones.

rtl urdu

Also noticed some additional issues that were also present before the switch to intl so I can make a new issue for those later.

chadian:
That's really interesting actually. I can't explain the font change for urdu. The characters look like they could be the same, I'm not sure if the change is acceptable or not either. I agree, let's do this as a separate issue and I'm happy to also investigate. Nothing else for this PR from my end, thanks! 👍

MatthewDorner:
The characters look to be identical, just different font so I doubt it's a big deal. Thanks for the quick fixes.

MatthewDorner:
Actually one more thing occurs to me, in https://github.com/HospitalRun/hospitalrun-frontend/pull/1458 we set all those translation keys to undefined instead of removing them so people could more easily see in the file what keys need to be translated.

Would it be better to comment them out instead of removing them entirely?

MatthewDorner:
Maybe that's not a good idea either since if people add keys, they'll just forget to add commented out keys in the other languages so they still won't stay synced up.

PR #1563, opened 2018-11-17
Fix 1561

### LINKED ISSUE #1561: ember-intl upgrade side effects

jackcmeyer:
Partial Work #1561

- fixes issue with warning messages rendering with raw HTML in the string instead of the formatted text

### DIFF
--- a/app/inventory/edit/template.hbs
+++ b/app/inventory/edit/template.hbs
@@ -10,7 +10,7 @@
           <button type="button" class="close" data-dismiss="alert"><span aria-hidden="true" class="octicon octicon-x"></span><span class="sr-only">{{t 'buttons.close'}}</span></button>
           <p>
             <span class="glyphicon glyphicon-warning-sign"></span>
-            {{t 'inventory.messages.quantity' quantity=model.quantity locationQuantity=locationQuantityTotal}}
+            {{t 'inventory.messages.quantity' quantity=model.quantity locationQuantity=locationQuantityTotal htmlSafe=true}}
           </p>
           <p>{{t 'inventory.messages.adjust' difference=quantityDifferential}}</p>
         </div>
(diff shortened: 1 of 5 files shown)


MatthewDorner:
We also need htmlSafe on the `dashboard.standalone_welcome` in index.hbs.

In some of those translation strings I see the {} is moved to a different position in the string from where it was before, like totalReceived, areYouSureDelete, maybe others. What was the reason for moving those?

jackcmeyer:
@MatthewDorner I've made the changes for the `dashboard.standalone_welcome` message. I've also removed the fixes for the warning messages because I think that I didn't do it right, even though google translate thought it was right. I will let someone who speaks or knows Hebrew make the changes.

MatthewDorner:
The addition of the parameters into the translation strings seemed good, I just wondered why the brackets changed place. It's OK since I noticed something else in the Hebrew file. For example, `incident.messages.incidentCategorySaved`. The translator translated the names of the parameters themselves, which results in the translations not working. And there are several instances of that in the Hebrew file, so we can make an issue to work on all that separately.

jackcmeyer:
They changed location because of the translations of the parameters being translated as well. So adding in the parameters will move the parentheses.

MatthewDorner:
I think I understand. Yeah that sounds difficult.

PR #1621, opened 2019-09-09
[WIP] Basic create react app

jackcmeyer:
Got some code for us? Awesome 🎊!

Please include a description of your change & check your PR against this list, thanks!

- basic create react app
- husky and commitizen
- commitlint

More info can be found by clicking the "guidelines for contributing" link above.


jackcmeyer:
@fox1t @tehKapa @PhearZero 

Here is a PR for a basic create react app. It has basic create react app, husky, commitzen, and and commitlint. This should address all tasks that were in the issue.

I have marked it as WIP since I have the following questions. 

- Does this repository need its own eslint/prettier config?
- Does this repository need .travis.yml enabled on it and its own .travis.yml config?
- Does this repository need its own semantic release process enabled?
- Should all of the .github templates be updated so that they can be standard across all repos? (Might be useful to create a template repo for the organization which standardizes formatting for issue/PR templates, README, contributing, etc.)

package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "hospitalrun-frontend",
```
PhearZero: I think we are going with **@hospitalrun-org** namespace. We can update this to `@hospitalrun-org/frontend`

package.json
```diff
@@ -0,0 +1,55 @@
...
+    ],
+    "development": [
+      "last 1 chrome version",
+      "last 1 firefox version",
+      "last 1 safari version"
+    ]
+  },
+  "husky": {
+    "hooks": {
+      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
```
PhearZero: Add a **pre-commit** hook that calls `npm test` for now

Example:
```
 "pre-commit": "npm test"
```

PhearZero:
> @jackcmeyer 
> 
> * Does this repository need its own eslint/prettier config?
> * Does this repository need .travis.yml enabled on it and its own .travis.yml config?
> * Does this repository need its own semantic release process enabled?
> * Should all of the .github templates be updated so that they can be standard across all repos? (Might be useful to create a template repo for the organization which standardizes formatting for issue/PR templates, README, contributing, etc.)

 -  Yes whatever CRA provides by default (react-app)
-  Yes but it is not a standard release cycle since we have dependent modules similar to Lerna projects. We do not want to release Feature X in the front end if Feature X is not implemented in the Server/Component libraries. This is done via the Mono-Repo HospitalRun/hospitalrun#17 HospitalRun/hospitalrun#16
-  Yes it will be triggered on Master from a Mono-Repo Build
-  I like the way you think, HospitalRun/hospitalrun#8 and the old [example in the PoC](https://github.com/HospitalRun/next-poc/tree/master/packages/hospital-run-docs) are going down that same path of thinking. We are going to have a standard storybook component `DocumentationPage` and CLI command called `hospitalrun docs` that will add any missing contribution guides, wrap readmes with headers and footers, and generate TSDocs to be used in Storybooks/wiki's etc.

package.json
jackcmeyer: addressed

package.json
jackcmeyer: addressed

jackcmeyer:
@PhearZero I can start working on getting a templates repository put together. I think it's probably outside the scope of this PR

PhearZero:
>  I can start working on getting a templates repository put together. I think it's probably outside the scope of this PR
@jackcmeyer Yea for sure lets head over to HospitalRun/hospitalrun#8 and I can get you up to date there

commitlint.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  extends: ["gitmoji"],
+  parserPreset: {
+      parserOpts: {
```
fox1t: @tehKapa is this compatible with the commit style you proposed?

package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@hospitalrun-org/frontend",
+  "version": "0.1.0",
+  "private": false,
+  "dependencies": {
+    "@types/jest": "24.0.18",
```
fox1t: Put all types in devDeps

package.json
```diff
@@ -0,0 +1,57 @@
...
+  },
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test",
+    "test:ci": "cross-env CI=true react-scripts test",
+    "eject": "react-scripts eject"
+  },
+  "eslintConfig": {
+    "extends": "react-app"
```
fox1t: Is this eslintConfig compatible with TS? I can't see the language plugins.

commitlint.config.js
tehKapa: theoretically, yes. however, I also extends this `['@commitlint/config-conventional']`
@PhearZero did you use both solutions in your tests?

fox1t: I really like conventional. In addition to it we need also to enforce referencing the issue the commits are addressing.

PhearZero: I started with config-conventional linting for testing locally and then added in alternative configurations in the PoC testing pnpm/husky.  We could rewrite the history if need be

@fox1t @tehKapa

package.json
PhearZero: It's what the CRA outputs with the --typescript flag. 
- [TypeScript CRA](https://create-react-app.dev/docs/adding-typescript)

PhearZero:
> @jackcmeyer we will use always the same tsconfig. This is our FE config https://github.com/HospitalRun/components/blob/master/tsconfig.json
> can you please ensure here are used the same compiler options?

Just a note react-scripts ignores some tsconfig options so we may have to eject if we want full control

commitlint.config.js
jackcmeyer: Is there anything that I need to do here?

This looks like it matches the config in the [hospitalrun/hospitalrun](https://github.com/HospitalRun/hospitalrun/blob/master/commitlint.config.js) repository.

If so, I would assume we would update this config to match [hospitalrun/components](https://github.com/HospitalRun/components/blob/master/commitlint.config.js)

jackcmeyer:
> @jackcmeyer we will use always the same tsconfig. This is our FE config [HospitalRun/components:tsconfig.json@`master`](https://github.com/HospitalRun/components/blob/master/tsconfig.json) [](/HospitalRun/components/blob/HEAD@{2019-09-09T08:24:21Z}/tsconfig.json)
> can you please ensure here are used the same compiler options?

I've updated the `tsconfig.json` to match the components library. However, I was not able to use declarations and isolatedModules together so I dropped declarations support. Let me know if that is ok

package.json
jackcmeyer: addressed

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM node:lts-alpine
```
tehKapa: @PhearZero usually I used this Dockerfile (docker in docker). what do you think?

```
FROM node:10-alpine as build

# set app basepath
ENV HOME=/home/app

# copy all app files
COPY . $HOME/node/

# change workgin dir and install deps in quiet mode
WORKDIR $HOME/node
RUN npm ci -q

# compile typescript and build all production stuff
RUN npm run build

# remove dev dependencies that are not needed in production
RUN npm prune --production

# start new image for lower size
FROM node:10-alpine

# create use with no permissions
RUN addgroup -g 101 -S app && adduser -u 100 -S -G app -s /bin/false app

# set app basepath
ENV HOME=/home/app

# copy production complied node app to the new image
COPY --from=build $HOME/node/ $HOME/node/
RUN chown -R app:app $HOME/*

# run app with low permissions level user
USER app
WORKDIR $HOME/node

EXPOSE 3000

ENV NODE_ENV=production

CMD [ "yarn", "start" ]
```

Issue #1612, opened 2019-07-15
Age  and DOB


joelelangovan1985:
**Expected
![age](https://user-images.githubusercontent.com/5588760/61896168-00505100-af32-11e9-8cc1-2f998c92f392.jpg)
 behavior:**
Many people in rural India, dont know the DOB, There should be a option to enter age instead of DOB
**Actual behavior:**
No Option to enter age

How do i overcome this problem ? age to DOB conversion would be useful.

JacobPie:
I feel like for a hospital, unfortunately, it needs to be more specific than just age. Maybe put in a general birthday?

joelelangovan1985:
Yes, but can there be an option to enter age which will convert like age to
standard 1-1-year like format.

tehKapa:
Hi @joelelangovan1985! 
We're [now working on a new major version](https://github.com/HospitalRun/hospitalrun-frontend/issues/1605#issuecomment-518749445) (2.0) of HospitalRun. This issue will be considered for the new release!

nicolasalarconrapela:
_Taking into account the above, he created a brief outline of how he could solve the issue of age and date of birth_:

![imagen](https://user-images.githubusercontent.com/23703592/62641303-97aaa080-b943-11e9-897a-05f54fa626b8.png)

_I explain briefly_:

The user not knowing the date of birth will **click on the checkbox of '`Unknown`'** , then :
* the field of '`Age`' will be **_unlocked_** and 
* the field of '`Date`' will be **_blocked_** and
    * **_`day`_** and **_`month`_** will be **_set to `01/01` automatically_** and
   * the _**`year` calculates based on age**_.

If the **'`Unknown`' checkbox is deselected** ( **BY DEFAULT**) : 
* the '`Age`' field will be blocked and calculated based on the date of birth entered and
* the '`Date`' field will be completely unlocked.

fox1t:
@nicolasalarconrapela Nice work, I think that this flow resolves this specific issue. We will add this to the official roadmap and features set.

jackcmeyer:
This functionality has been implemented in v2 (see #1672). Closing this issue.

PR #1672, opened 2019-12-18
feat(patients): add ability to create new patient

### LINKED ISSUE #1671: Create New Patient
## Scenario
**AS** a user with patient write access
**I WANT** to be able to create a new patient
**SO THAT** I can enter a new patient into my system

## Acceptance Criteria
**GIVEN** there is a new patient to be added to the system
**WHEN** I access the /new/patients page
**THEN** I should see a form that allows me to enter the following data: 

* Name - (prefix, given name, family name, suffix)
* Sex
* Date of Birth
* Approximate Age
* Patient Type
* Occupation
* Preferred Language
* Phone Number
* Email
* Address

**GIVEN** a new patient form
**WHEN** I select the approximate age checkbox
**THEN** A new text box appears that allows me to enter an approximate age
**AND THEN** when I enter the approximate age, the date of birth field should be populated with the date which equals current date - approximate age.

## Technical Notes
https://www.hl7.org/fhir/patient.html

## Related Issues
* #1612

jackcmeyer:
Fixes #1671 #1612 .

- adds ability to create a new patient
- adds ability to approximate age

- [date-fns](https://bundlephobia.com/result?p=date-fns@2.8.1)


src/components/Navbar.tsx
```diff
@@ -32,7 +31,7 @@ const Navbar = () => {
             {
               label: t('actions.new'),
               onClick: () => {
-                history.push('/patients/new')
+                history.push('/new/patients')
```
fox1t: Why changing this line? I think it is better to got with `/entityName(s)/new` for every entity we have

src/model/Patient.ts
```diff
@@ -1,13 +1,51 @@
...
+  preferredLanguage?: string
+
+  occupation?: string
+
+  type?: string
+
+  constructor(
+    id: string,
+    rev: string,
+    name: Name,
```
fox1t: I am not sure we need a custom class for name. Another problem in having 2 depth prop is that we have to check if `.name` exists before doing `.name.given`. I think it is better to flat it.

src/components/input/DatePickerWithLabelFormGroup.tsx
```diff
@@ -0,0 +1,30 @@
...
+  value: string
+  isEditable: boolean
+  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void
+}
+
+const DatePickerWithLabelFormGroup = (props: Props) => {
+  const { onChange, label, name, value, isEditable } = props
+  const id = `${name}DatePicker`
+  return (
+    <div className="form-group">
```
fox1t: Why not using our date picker component here?

src/model/Name.ts
```diff
@@ -0,0 +1,16 @@
+export default class Name {
```
fox1t: Not sure we need this class.

fox1t:
This PR is not passing tests at the moment: `> 47 |     history.push(`patients/${newPatient.id}`)` produces `Attempted to log "TypeError: Cannot read property 'body' of null`

package.json
```diff
@@ -10,10 +10,12 @@
     "@semantic-release/changelog": "~3.0.4",
     "@semantic-release/git": "~7.0.16",
     "@semantic-release/release-notes-generator": "~7.3.0",
+    "@types/luxon": "^1.21.0",
     "bootstrap": "~4.4.1",
     "i18next": "^19.0.1",
     "i18next-browser-languagedetector": "^4.0.1",
     "i18next-xhr-backend": "^3.2.2",
+    "luxon": "^1.21.3",
```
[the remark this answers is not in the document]
jackcmeyer: For no good reason other than familiarity with luxon/moment. 

I can investigate using `date-fns`

src/components/Navbar.tsx
jackcmeyer: I'm considered that if we have route such as 

`/entity/:id`
and 
`/entity/new`

That we could end up with collisions. 

This guarantees that we will never have that happen.

src/components/input/DatePickerWithLabelFormGroup.tsx
jackcmeyer: updated to use our date picker

src/components/Navbar.tsx
jackcmeyer: I updated the route back to what it was. 😊

src/model/Name.ts
jackcmeyer: Removed this class.

src/model/Patient.ts
jackcmeyer: Changed it to flatten the name object.

package.json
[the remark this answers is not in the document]
jackcmeyer: Switched to use date-fns. First time using it, I enjoyed their API. 🎉

PR #1736, opened 2020-01-18
feat:error handling to patient form

M-BenAli:

- error handling to patient form so users can't submit forms without a patient name
- added .idea/ default folder to .gitignore for intellij IDE users(like myself)


src/patients/new/NewPatientForm.tsx
```diff
@@ -274,7 +278,9 @@ const NewPatientForm = (props: Props) => {
             />
           </div>
         </div>
-
+        {errorMessage && (
+          <div className="alert alert-danger" role="alert">{t(errorMessage)}</div>
```
jackcmeyer: We can use the `Alert` component from `@hospitalrun/component`.

src/patients/new/NewPatientForm.tsx
```diff
@@ -37,6 +38,9 @@ const NewPatientForm = (props: Props) => {
   })
 
   const onSaveButtonClick = async () => {
+    if (!patient.givenName && !patient.familyName) {
```
jackcmeyer: In order to support all different types of name formats, I think that we should require that ONE of these fields is present (doesn't matter which one). @fox1t @StefanoMiC @tehKapa @lauggh what do you all think here?

src/patients/new/NewPatientForm.tsx
```diff
@@ -19,6 +19,7 @@ const NewPatientForm = (props: Props) => {
   const [isEditable] = useState(true)
   const { onCancel, onSave } = props
   const [approximateAge, setApproximateAge] = useState(0)
+  const [errorMessage, setError] = useState('')
```
jackcmeyer: Rename `setError` to `setErrorMesssage` to keep the state variable and state update function consistent.

src/patients/new/NewPatientForm.tsx
```diff
@@ -37,6 +38,9 @@ const NewPatientForm = (props: Props) => {
   })
 
   const onSaveButtonClick = async () => {
+    if (!patient.givenName && !patient.familyName) {
+      return setError("No patient name entered!")
```
jackcmeyer: This string needs to support internationalization. 

To do that you can add a key and value to `locales/en-US/translation.json`. I think a key in the `patient` object called `patientNameRequired` or something like that would work.

src/patients/new/NewPatientForm.tsx
jackcmeyer: the value passed to the `t` function needs to exist in the translations file.

jackcmeyer: I think that error messages should be above the form fields. @lauggh @StefanoMiC thoughts here?

jackcmeyer:
Thanks for the contribution @M-BenAli!

A few things on this PR: 
- Left some feedback on the code
- Some linting errors causing the build to fail
- Need some tests surrounding this new functionality
- Please make sure to sign the [CLA](https://cla.js.foundation/HospitalRun/hospitalrun-frontend?pullRequest=1736) with the account that you committed with.

jackcmeyer:
It looks like the account you committed with is not a GitHub user. Either you'll need to create an account for those commits, or you can amend your commits to match the account you opened the pull request with: https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-one-specific-commit/28845565

jackcmeyer:
Also, congratulations on your first ever pull request @M-BenAli 🎉🎉🎉!!

src/patients/new/NewPatientForm.tsx
fox1t: I agree with you @jackcmeyer. For example we already found cases where familiy names are not used (Burkina Faso). I think that the best choice is to make this configurable: the admin can decide mandatory fields for "all" forms. What do you think?

src/patients/new/NewPatientForm.tsx
fox1t: +1

fox1t:
🙏 Thanks for your contribution! We will guide you to make this PR merged! ☺️

M-BenAli:
Hi all! Thank you for the feedback and the welcome messages, I'll be looking at the requested changes and adjust the code.

src/patients/new/NewPatientForm.tsx
```diff
@@ -93,6 +98,7 @@ const NewPatientForm = (props: Props) => {
     <div>
       <form>
         <h3>{t('patient.basicInformation')}</h3>
+        {errorMessage && <Alert className="alert" color="danger" message={t(errorMessage)} />}
```
jackcmeyer: No need for the call to `t` here. We can simply just pass error message to message since the error message was already internationalized.

src/locales/en-US/translation.json
```diff
@@ -52,6 +52,9 @@
   "states": {
     "success": "Success!"
   },
+  "errors": {
+    "patientNameRequired": "No patient name entered!"
```
jackcmeyer: To aid in finding translations easier, I think a better spot for this error message would be in the `patients` object. 

So the file would end up something like
```json
{
  ...
  "patient": {
	...,
	"errors": {
		"patientGivenNameRequired": "Patient Given Name is required."
	}
  }
}
```

jackcmeyer: I would suggest `Patient Given Name is required.` for the error message.

M-BenAli:
> Everything is starting to look really good! Now just need some tests covering this new functionality.

I'll be covering the tests later this week

src/__tests__/patients/new/NewPatient.test.tsx
```diff
@@ -1,11 +1,11 @@
 import '../../../__mocks__/matchMediaMock'
 import React from 'react'
-import { mount } from 'enzyme'
-import { Router, MemoryRouter } from 'react-router-dom'
-import { Provider } from 'react-redux'
-import { mocked } from 'ts-jest/utils'
-import { createMemoryHistory } from 'history'
-import { act } from 'react-dom/test-utils'
```
fox1t: Have our prettier rules changed? Why these spaces are removed?

M-BenAli: I'll look into it, it could be that my IDE is ignoring the eslint rules when formatting the code.

tehKapa: What IDE do you use?

M-BenAli: > What IDE do you use?

Intellij idea ultimate

jackcmeyer: Running `yarn lint` should automatically format. 

@fox1t @tehKapa we may want to put this in a pre commit hook.

jackcmeyer: Also, we need to investigate why this isn't failing the build.

fox1t: @tehKapa do you have any ideas why this is not failing or automatically formatting?

tehKapa: mmmh no

M-BenAli:
> The only question I have is: why all surrounding spaces are removed in `{}` and `< />`?

This problem should be solved now in the latest commit, `npm run lint` did the trick for me(using npm instead of yarn).

fox1t:
> > The only question I have is: why all surrounding spaces are removed in `{}` and `< />`?
> 
> This problem should be solved now in the latest commit, `npm run lint` did the trick for me(using npm instead of yarn).

We need to investigate into this. Thanks for the feedback!

fox1t:
I opened a new regression issue: https://github.com/HospitalRun/hospitalrun-frontend/issues/1775

fox1t:
LGTM

PR #1738, opened 2020-01-20
feat(patients): add related persons

### LINKED ISSUE #1689: Add relationships to patient
## Scenario
**AS**

A user with write access to patient data

**I WANT**

The ability to add relationships to a patient

**SO THAT**

I can easily understand the relationships that a patient has.

## Acceptance Criteria
**GIVEN**

A logged in user with write access to patient data and an existing patient

**WHEN**

I access the `/patients/:id` route and click the add relationship button

**THEN**

A modal appears with a form that has the ability to enter relationship information.

Information to collect: 
* type of relationship
* prefix, given, family, and suffix names for the related person
* phone number of the related person
* address of the related person

## Related Issues
* HospitalRun/hospitalrun#10

## Technical Notes
* https://www.hl7.org/fhir/relatedperson.html
* Types of related persons: https://www.hl7.org/fhir/valueset-relatedperson-relationshiptype.html

jackcmeyer:
Fixes #1689 .

- adds the ability to define a related person to the given patient
- refactored some models to eliminate the need to duplicate code between the related person and patient model.


fox1t:
@jackcmeyer is it possible, in this phase of development, to allow adding a patient as relative? It is important to NOT repeat 2 times a person, if we already have it as patient in our systems.

jackcmeyer:
@fox1t yes we would be able to do that now. We have typeahead functionality and patient search functionality which I think is all that is needed. 

I was considering doing this in this implementation before, but decided that we might want to open a separate issue to do it. 

Thoughts?

fox1t:
I think that we can handle it here and it would be amazing. I think that this is one of most underestimated features.

jackcmeyer:
@fox1t sounds good! I’ll get started on it tonight!

jackcmeyer:
@fox1t @tehKapa this PR requires: https://github.com/HospitalRun/components/pull/236 to be merge and then this branch updated to use the new version.

src/clients/db/Repository.ts
```diff
@@ -52,6 +52,27 @@ export default class Repository<T extends AbstractDBModel> {
...
+      const entityToUpdate = {
+        _id: id,
+        _rev: rev,
+        ...restOfDoc,
+        ...entity,
+      }
+      await this.db.put(entityToUpdate)
+      return this.find(entity.id)
+    } catch (error) {
+      return this.save(entity)
```
fox1t: Can you explain why saving after an error?

src/model/Name.ts
```diff
@@ -0,0 +1,7 @@
+export default interface Name {
+  prefix?: string
+  givenName?: string
+  familyName?: string
+  suffix?: string
+  fullName?: string
```
fox1t: I am still not convinced having fullName as field: it could be very very dangerous into the wild.

src/clients/db/Repository.ts
jackcmeyer: So this function is for save and update functionality. We want to save a NEW entity if it didn't' exist in the database and we want to UPDATE an existing entity if it did. 

So the first thing that it does is it tries to find an existing entity. If it didn't find the entity, then it throws an error, which lets us know that we should save a new entity.

fox1t:
LGTM

PR #1780, opened 2020-02-02
fix(build): fix precommit lint hook issue

### LINKED ISSUE #1775: Lint command is not fixing properly the code
## 💥 Regression Report

In this [PR](https://github.com/HospitalRun/hospitalrun-frontend/pull/1736) we discovered that the code wasn't formatted properly. 

## Last working version

Worked up to version: We don't really know

## To Reproduce

* Commit some code that is clearly not properly formatted, ex. `import {cuzz} from 'cuzzo'` instead of the correct`import { cuzz } from 'cuzzo'`.

## Expected behavior
In the above example the code should be formatted and committed correctly. 
```typescript
`import { cuzz } from 'cuzzo'`
```

ocBruno:
fix #1775

- added pre-commit hook and fix lint-staged config
```
"husky": {
    "hooks": {
      "pre-commit": "lint-staged && yarn test:ci",
      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
    }
  },
"lint-staged": {
    "*.{js,jsx,ts,tsx}": [
      "eslint --fix"
    ],
    "*.js": "eslint --cache --fix"
  }
```


jackcmeyer:
Adding the pre-commit hook is a good idea, @ocBruno!

However, I'm not sure that this fixes the initial issue. What was happening is that the build was not failing if there was lint issues because `yarn lint` runs `eslint .... --fix` during the build so any fixable lint issues are not reported on.

There are ways around precommit hooks like using `--no-verify` that would still allow for lint errors to get committed and pushed. 

I think that during the build we need to just run `eslint ...` (without the fix) to ensure builds fail when fixable lint issues are committed and pushed. 

@tehKapa @fox1t what do you think?

ocBruno:
> Adding the pre-commit hook is a good idea, @ocBruno!
> 
> However, I'm not sure that this fixes the initial issue. What was happening is that the build was not failing if there was lint issues because `yarn lint` runs `eslint .... --fix` during the build so any fixable lint issues are not reported on.
> 
> There are ways around precommit hooks like using `--no-verify` that would still allow for lint errors to get committed and pushed.
> 
> I think that during the build we need to just run `eslint ...` (without the fix) to ensure builds fail when fixable lint issues are committed and pushed.
> 
> @tehKapa @fox1t what do you think?



> Adding the pre-commit hook is a good idea, @ocBruno!
> 
> However, I'm not sure that this fixes the initial issue. What was happening is that the build was not failing if there was lint issues because `yarn lint` runs `eslint .... --fix` during the build so any fixable lint issues are not reported on.
> 
> There are ways around precommit hooks like using `--no-verify` that would still allow for lint errors to get committed and pushed.
> 
> I think that during the build we need to just run `eslint ...` (without the fix) to ensure builds fail when fixable lint issues are committed and pushed.
> 
> @tehKapa @fox1t what do you think?

I was removing the --fix option and ran into another issue with eslint and prettier config. 
I was receiving errors on every line saying i needed to remove CR so I did some troubleshooting and realized some misisng configs.
Eslint-config-prettier was present in the packages but not in the last place of the eslint extends array as presented below.
https://prettier.io/docs/en/integrating-with-linters.html
Prettier was missing the rules array as well so I added it. 
I think this resolves some build issues but not sure if the structure is correct.

package.json
```diff
@@ -120,6 +120,7 @@
   },
   "husky": {
     "hooks": {
+      "pre-commit": "lint-staged && yarn test && yarn test:ci",
```
jackcmeyer: I don't think that we need to run `yarn test && yarn test:ci`. `yarn test:ci` should be sufficient

.prettierrc
```diff
@@ -6,5 +6,6 @@
   "trailingComma": "all",
   "bracketSpacing": true,
   "jsxBracketSameLine": false,
-  "arrowParens": "always"
-}
\ No newline at end of file
+  "arrowParens": "always",
+  "endOfLine": "auto"
```
jackcmeyer: I think that we want `"endOfLine"` to be `LF` to be consistent with unix machines

src/components/Navbar.tsx
```diff
@@ -6,7 +6,6 @@ import { useTranslation } from 'react-i18next'
 const Navbar = () => {
   const { t } = useTranslation()
   const history = useHistory()
```
jackcmeyer: interesting that the whitespace that got remove din this file...

package.json
```diff
@@ -129,10 +130,9 @@
...
   "lint-staged": {
-    "**/*.{js,jsx,ts,tsx}": [
-      "npm run lint",
-      "npm run test:ci",
-      "git add ."
-    ]
+    "*.{js,jsx,ts,tsx}": [
+      "eslint"
+    ],
+    "*.js": "eslint --cache"
```
jackcmeyer: Is it possible for this config to be equal to the one from the components repo

```json
  "lint-staged": {
    "**/*.{js,jsx,ts,tsx}": [
      "npm run lint",
      "git add ."
    ]
  },
```

src/components/Navbar.tsx
```diff
@@ -6,7 +6,6 @@ import { useTranslation } from 'react-i18next'
...
-
```
jackcmeyer: weird that whitespace got removed from this file.

package.json
jackcmeyer: Is it possible for this to be the same config as the components repository: 

```json
  "lint-staged": {
    "**/*.{js,jsx,ts,tsx}": [
      "npm run lint",
      "git add ."
    ]
  },
```

.prettierrc
fox1t: I agree, we absolutely need consistent endOfLine. You can see this setting also here: https://github.com/HospitalRun/hospitalrun-frontend/blob/master/.editorconfig#L9

package.json
fox1t: @tehKapa I want you thoughts on this.

ocBruno: I also think it should be applied to components library, not sure if that period in git add . is redundant and if that cached lint is worth it but I do believe we should add a pre commit test to the components project

fox1t:
> Adding the pre-commit hook is a good idea, @ocBruno!
> 
> However, I'm not sure that this fixes the initial issue. What was happening is that the build was not failing if there was lint issues because `yarn lint` runs `eslint .... --fix` during the build so any fixable lint issues are not reported on.
> 
> There are ways around precommit hooks like using `--no-verify` that would still allow for lint errors to get committed and pushed.
> 
> I think that during the build we need to just run `eslint ...` (without the fix) to ensure builds fail when fixable lint issues are committed and pushed.
> 
> @tehKapa @fox1t what do you think?

@jackcmeyer you are write. Removing --fix ONLY on CI pipeline might be the correct fix.

.eslintrc.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
     'prettier',
     'prettier/@typescript-eslint',
     'plugin:prettier/recommended',
+    'eslint-config-prettier'
```
tehKapa: We run prettier as an eslint plugin, so this harmful

.eslintrc.js
```diff
@@ -31,6 +32,7 @@ module.exports = {
   },
   plugins: ['react', '@typescript-eslint', 'prettier', 'jest'],
   rules: {
+    "prettier/prettier": "error",
```
tehKapa: `'plugin:prettier/recommended'` already includes this rule. Refence: https://github.com/prettier/eslint-plugin-prettier#recommended-configuration

.prettierrc
```diff
@@ -6,5 +6,6 @@
   "trailingComma": "all",
   "bracketSpacing": true,
   "jsxBracketSameLine": false,
-  "arrowParens": "always"
-}
\ No newline at end of file
+  "arrowParens": "always",
+  "endOfLine": "lf"
```
tehKapa: I force this to lf

package.json
```diff
@@ -129,11 +130,10 @@
       "path": "./node_modules/cz-conventional-changelog"
     }
   },
-  "lint-staged": {
+    "lint-staged": {
     "**/*.{js,jsx,ts,tsx}": [
       "npm run lint",
-      "npm run test:ci",
```
tehKapa: it's important to test commit before add files to stage

package.json
```diff
@@ -121,6 +121,7 @@
   },
   "husky": {
     "hooks": {
+      "pre-commit": "lint-staged && yarn test:ci",
```
tehKapa: see the next comment (remove `&& yarn test:ci`)

ocBruno: Applied the changs but am receiving this error on commit.
```

× npm found some errors. Please fix them and try committing again.

> @hospitalrun/frontend@0.0.0-development test:ci C:\Users\55419\Code\hospitalrun-frontend
> cross-env CI=true react-scripts test "C:/Users/55419/Code/hospitalrun-frontend/.eslintrc.js"

No tests found, exiting with code 1
Run with `--passWithNoTests` to exit with code 0
In C:\Users\55419\Code\hospitalrun-frontend
82 files checked.
testMatch: C:/Users/55419/Code/hospitalrun-frontend/src/**/__tests__/**/*.{js,jsx,ts,tsx}, C:/Users/55419/Code/hospitalrun-frontend/src/**/*.{spec,test}.{js,jsx,ts,tsx} - 30 matches
testPathIgnorePatterns: \
ode_modules\\ - 82 matches
testRegex:  - 0 matches
Pattern: C:\\Users\\55419\\Code\\hospitalrun-frontend\\.eslintrc.js - 0 matches
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @hospitalrun/frontend@0.0.0-development test:ci: `cross-env CI=true react-scripts test "C:/Users/55419/Code/hospitalrun-frontend/.eslintrc.js"`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the @hospitalrun/frontend@0.0.0-development test:ci script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\55419\AppData\Roaming
pm-cache\_logs\2020-02-07T13_30_53_752Z-debug.log
husky > pre-commit hook failed (add --no-verify to bypass
```

tehKapa: I [pushed](https://github.com/HospitalRun/hospitalrun-frontend/pull/1780/commits/639f122b11e6eee2953d0ad40db6edbd22b82f61) changes myself and didn't have any trouble.

PR #1788, opened 2020-02-04
fix(patients): fix issue with patient duplicating on update

### LINKED ISSUE #1783: Updating patient causes duplicate due to document version conflict
## 🐛 Bug Report
When you update an existing document in PouchDB, you need to give the database the `_rev` of the document that's currently in the database. If you supply an older `_rev`, it will cause a document update conflict. When we update a Patient, this creates a new `_rev` but we are not pulling that out of the database and updating our Redux state with it. The logic in `saveOrUpdate` currently falls back to a regular save when an error is thrown from `db.put`, so if you try to update a Patient twice it creates a duplicate document (the first update works).

## To Reproduce
Create a Patient and add two Related Persons. The Patient will be duplicated.

jackcmeyer:
Fixes #1783 

- fixes issue with wrong revision number in the save or update method
- fixes issue where new related person was being added to a immutable array


tehKapa:
There seems to be a bug with node 10 :|

PR #1792, opened 2020-02-05
feat(navigation): navigate to patients profile on related person click

### LINKED ISSUE #1763: Clicking on related person should navigate to patient profile
## Scenario
**AS**

A user who has read access to patients

**I WANT**

The ability to click on a related person and have it bring me to their patient profile

**SO THAT**

I can view information about that related peron

## Acceptance Criteria
**GIVEN**

An existing patient

AND

An existing related person

AND 

Read access to patiens
**WHEN**

I click on the related person 
**THEN**

The application should navigate the to that related person's patient profile at `/patients/:id`

ocBruno:
Fixes #1763

- clicking on related person navigates to patient profile
- add history to related person tests and refactor the tests to fix conflicts


src/__tests__/patients/related-persons/RelatedPersons.test.tsx
```diff
@@ -78,34 +86,36 @@ describe('Related Persons Tab', () => {
...
-      })
-
-      wrapper.update()
-
-      act(() => {
-        const newRelatedPersonModal = wrapper.find(NewRelatedPersonModal)
-
-        const onSave = newRelatedPersonModal.prop('onSave') as any
-        onSave(expectedRelatedPerson)
+      await act(async () => {
```
jackcmeyer: What is the reason for this test getting updated?

src/__tests__/patients/related-persons/RelatedPersons.test.tsx
```diff
@@ -1,5 +1,9 @@
 import '../../../__mocks__/matchMediaMock'
 import React from 'react'
+import { Router } from 'react-router'
+
+import { createMemoryHistory } from 'history'
+
```
jackcmeyer: Some extra white space at the top of this file

src/__tests__/patients/related-persons/RelatedPersons.test.tsx
```diff
@@ -18,12 +22,23 @@ const mockStore = createMockStore([thunk])
...
+    history = createMemoryHistory()
+    history.push(location)
+    return mount(
+      <Router history={history}>
+        <Provider store={mockStore({ patient, user })}>
+          <RelatedPersonTab patient={patient} />
+        </Provider>
+      </Router>,
+    )
+  }
```
jackcmeyer: Not sure if we need a setup method in this test. 

I think that doing the setup in a `beforeEach` is sufficient. Thoughts?

src/__tests__/patients/related-persons/RelatedPersons.test.tsx
ocBruno: I was getting errors and there was an async operation that used the wrappers which I was creating with the setup function so I made setup async and adapted the rest of the file. The test is indeed missing the new functionality, I refactored all the tests for the changes but forgot to add the new one. If it would be more performant and reusable we make the beforeeach but I think we with the setup wrapper we can call it once on the describe scope instead of each test and in cases where we need a new patient user or route we call the setup function in the individual test. Give me a heads up and I'll finish refactoring this as necessary

src/__tests__/patients/related-persons/RelatedPersons.test.tsx
ocBruno: If it would be more performant and reusable we make the beforeeach but I think we with the setup wrapper we can call it once on the describe scope instead of each test and in cases where we need a new patient user or route we call the setup function in the individual test. Give me a heads up and I'll finish refactoring this as necessary

jackcmeyer: I think a setup may not be necessary because each describe has a different use case. 

I’ve found the setup function to be helpful when the same describe needs different setup.

ocBruno: Reverted the test changes and added the related test!

PR #1810, opened 2020-02-06
fix(patients): stop "Loading..." when patient has no related persons

### LINKED ISSUE #1789: Related Persons screen always shows loading when there are no related persons
🐛 Bug Report

When a patient does not have a related person, the related person tab always shows loading. 

## To Reproduce

1. Navigate to a patient without a related person
2. Navigate to the related persons tab
3. See that the related person tab always shows loading

## Expected behavior
A warning message should appear stating that there are no related persons and describe how to add one. 

However, it should show a loading message when it is loading.

archwheeler:
Fixes #1789.

- Display something more useful than "Loading..." when a patient has no related persons:

related


src/patients/related-persons/RelatedPersonTab.tsx
```diff
@@ -88,14 +88,17 @@ const RelatedPersonTab = (props: Props) => {
...
-              <h1>Loading...</h1>
-            )}
+              (relatedPersons.length > 0) ? (
+                <List>
+                  {relatedPersons.map((r) => (
+                    <ListItem key={r.id}>{r.fullName}</ListItem>
+                  ))}
+                </List>
+                ) : (
+                  <h2>No related persons have been added yet.</h2>
```
jackcmeyer: I think we can use the alert component here, using the warning color (or maybe info)

src/patients/related-persons/RelatedPersonTab.tsx
```diff
@@ -88,14 +88,17 @@ const RelatedPersonTab = (props: Props) => {
         <div className="col-md-12">
           <Panel title={t('patient.relatedPersons.label')} color="primary" collapsible>
             {relatedPersons ? (
-              <List>
-                {relatedPersons.map((r) => (
-                  <ListItem key={r.id}>{r.fullName}</ListItem>
-                ))}
-              </List>
-            ) : (
-              <h1>Loading...</h1>
```
jackcmeyer: We still should display loading if the page is actually loading. 

I think we can add a `isLoading` state and set it to `true` right before fetching the related persons and then false when its finished`

jackcmeyer:
Congrats on your first PR to the project 🥇

src/patients/related-persons/RelatedPersonTab.tsx
archwheeler: Sounds good! I've made another commit, is this any better?

related

src/patients/related-persons/RelatedPersonTab.tsx
archwheeler: The loading message should still be displayed if the page is loading:

"Loading..." is displayed when `relatedPersons` is undefined, which it initially is when the page starts loading.

The bug was to do with the `setRelatedPersons()` call being inside `if (patient.relatedPersons)`, resulting in `relatedPersons` being undefined instead of empty if a patient has no related persons.

My commit has moved the `setRelatedPersons` call so that it is called even when a patient has no related persons (setting `relatedPersons` to an empty array).

Hopefully this solution is okay. If needed I could commit a different solution using a new `isLoading` state.

fox1t:
LGTM

src/patients/related-persons/RelatedPersonTab.tsx
jackcmeyer: Yeah, I think your changes definitely fixed the bug! I think we now need to add in the intended functionality of displaying a loading message while the page is loading, a warning message when there are no related persons, and the related persons list when there are related persons.

src/patients/related-persons/RelatedPersonTab.tsx
```diff
@@ -88,14 +88,17 @@ const RelatedPersonTab = (props: Props) => {
...
-            )}
+              (relatedPersons.length > 0) ? (
+                <List>
+                  {relatedPersons.map((r) => (
+                    <ListItem key={r.id}>{r.fullName}</ListItem>
+                  ))}
+                </List>
+                ) : (
+                  <Alert color="warning" title="No Related Persons" message="Add a related person using the button above." />
```
jackcmeyer: The title and message need to be internationalized.

See: https://github.com/HospitalRun/hospitalrun-frontend/blob/master/src/locales/en-US/translation.json

src/patients/related-persons/RelatedPersonTab.tsx
archwheeler: I believe my changes already display the loading message whilst loading, the warning message when there are no related persons or the related persons list when there are related persons?

src/patients/related-persons/RelatedPersonTab.tsx
archwheeler: I've internationalized the Related Persons tab now, hopefully I've done it correctly!

src/__tests__/patients/related-persons/RelatedPersons.test.tsx
```diff
@@ -196,4 +196,39 @@ describe('Related Persons Tab', () => {
...
+      })
+      wrapper.update()
+    })
+
+    it('should display a warning if patient has no related persons', () => {
+      const warning = wrapper.find(Alert)
+      expect(warning).toBeDefined()
+    })
+  })
 })
```
archwheeler: I've also added a test to ensure the warning message is displayed when a patient has no related persons.

src/patients/related-persons/RelatedPersonTab.tsx
jackcmeyer: You're right! My bad, I saw the deleted line in the diff and then didn't see the addition in a different spot!

I think the only thing to change is instead of the `Loading...` text, we could use the `Spinner` component

src/patients/related-persons/RelatedPersonTab.tsx
jackcmeyer: you got it! nice work

src/patients/related-persons/RelatedPersonTab.tsx
archwheeler: I've replaced the `Loading...` text with the `Spinner` component in a new commit. :ok_hand:

PR #1815, opened 2020-02-09
feat(breadcrumb): add a breadcrumb underneath the page header

### LINKED ISSUE #1770: Add breadcrumbs
## Scenario
**AS**
As a user

**I WANT**

Breadcrumbs on the page

**SO THAT**

I can have easy navigation capabilities.

## Acceptance Criteria
**GIVEN**

A logged in user

**WHEN**

I view the application 

**THEN**

I see a breadcrumb underneath the page header

and when I navigate between pages the breadcrumb updates

and there are links to the previous pages in the breadcrumb.

oliv37:
Fixes #1770.

- Add a breadcrumb underneath the page header
- The breadcrumb is updated when the user navigates between pages

Let me know if the breadcrumb created in this PR is what you are expecting, then I will finish all the unit tests


jackcmeyer:
UI Style Feedback: I think that the breadcrumbs should go directly below the page break line, rather than right below the Navbar.

src/components/breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,16 @@
...
+import AppointmentBreadcrumb from 'components/breadcrumb/AppointmentBreadcrumb'
+
+const Breadcrumb = () => (
+  <Switch>
+    <Route exact path={['/patients/new', '/appointments/new']} component={DefaultBreadcrumb} />
+    <Route path="/patients/:id" component={PatientBreadcrumb} />
+    <Route path="/appointments/:id" component={AppointmentBreadcrumb} />
+    <Route path="*" component={DefaultBreadcrumb} />
+  </Switch>
+)
```
jackcmeyer: I'm wondering if we can combine all of the breadcrumbs into one component. 

This breadcrumb component would read from a Provider (defined using the React Context API) or a Redux Slice. 

The state the Provider or Slice would be an array of objects that have two properties: text and  location. 

The Provider or Slice would have three functions exposed, one to add a breadcrumb, one to remove a breadcrumb, and one to set the breadcrumbs (basically a function to just set the state exactly to what the user passed in). 

Each Route Handler (e.g. `Patients`, `ViewPatient`, `NewPatient`) would then add a breadcrumb to the state, and would remove it during the cleanup phase of the `useEffect`.

This would ensure we don't have any duplicate code, the breadcrumb can be a single component, and we don't have any additional work each time a new route or module is added.

Feel free to ping me on slack with any questions.

jackcmeyer: For an example of how this might work, checkout https://github.com/HospitalRun/hospitalrun-frontend/pull/1817/files. I'm working through doing something similar but with a button toolbar.

oliv37: Ok, if my understanding is correct, each Route Component will declare an `effect` to update the breadcrumb array in the state ?

jackcmeyer: I think that you have the correct understanding!

oliv37: I updated my code using your recommendations (without doing the unit tests).
I only handled the `/` and `/patients/**` routes, let met know if it's ok for you, then I will finish the tests and handle the `/appointments/**` routes

package.json
```diff
@@ -127,7 +127,6 @@
   "lint-staged": {
     "**/*.{js,jsx,ts,tsx}": [
       "npm run lint:fix",
-      "npm run test:ci",
```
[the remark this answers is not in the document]
oliv37: yes I know, I deleted it to push my commits (without doing the tests). I will put it back in my next commit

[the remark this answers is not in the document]
fox1t: OK i think you can put it back so we can run tests :)
Good job!

src/components/breadcrumb/Breadcrumb.tsx
oliv37: I just finished to implement the whole feature (I still need to improve the code coverage). 

It was not as easy as I expected because of the nested routes. I realized the effect of a child component is executed before the effect of its parent. As a result I couldn't rely on the order of the addBreadcrumb function calls to build a valid breadcrumbs array. So I decided to sort the breadcrumbs by their location property before the rendering. 

I am also afraid that maintaining the breadcrumbs will be painful in the future, as the number of routes will increase. Each developers should remember that when they add a route, they should also call the hook `useAddBreadcrumbs`, it can be really cumbersome...

tehKapa:
@lauggh would a functional UX require a "Dashboard" (sort of like a homepage) clickable before "Patients" item? 
Screenshot 2020-02-18 at 22 22 42

lauggh:
Yes, the "Dashboard" in the breadcrumbs should be linked.

tehKapa:
@oliv37 can you add this change?

oliv37:
> @lauggh would a functional UX require a "Dashboard" (sort of like a homepage) clickable before "Patients" item?

Do I also need to a the "Dashboard" (sort of link to the homepage) before "Appointments" item ?

tehKapa:
@lauggh what do you say? Another note... does it make sense to have breadcrumbs in the Dashboard? 
image

lauggh:
No breadcrumbs on the top level. Good catch

https://github.com/HospitalRun/hospitalrun-frontend/pull/1815#issuecomment-588432642

oliv37:
Still no answer to my first question, Do I need to add "Dashboard" at the beginning of all breadcrumbs (/patients/**, /appointments/**) ?

lauggh:
Yes add Dashboard on all breadcrumbs as a link except on the Dashboard page
itself

oliv37:
it might be ok now

tehKapa:
@oliv37 can you resolve conflicts? :) thx

PR #1823, opened 2020-02-13
feat(AppointmentsList): add an appointments tab to the patient view

igeagonz:
add an appointments tab to the patient view which shows appointments list for corresponding patient.
Uses same layout as patient list view and also implements fuzzy search

#1769  

- add new appointments tab
- allow fuzzy search

### DIFF
--- a/src/clients/db/AppointmentsRepository.ts
+++ b/src/clients/db/AppointmentsRepository.ts
@@ -6,6 +6,38 @@ export class AppointmentRepository extends Repository<Appointment> {
   constructor() {
     super(appointments)
   }
+
+  // Fuzzy search for patient appointments. Used for patient appointment search bar
+  async searchPatientAppointments(patientId: string, text: string): Promise<Appointment[]> {
+    return super.search({
+      selector: {
+        $and: [
+          {
+            patientId,
+          },
+          {
+            $or: [
+              {
+                location: {
+                  $regex: RegExp(text, 'i'),
+                },
+              },
+              {
+                reason: {
+                  $regex: RegExp(text, 'i'),
+                },
+              },
+              {
+                type: {
+                  $regex: RegExp(text, 'i'),
+                },
+              },
+            ],
+          },
+        ],
+      },
+    })
+  }
 }
 
 export default new AppointmentRepository()
--- a/src/patients/appointments/AppointmentsList.tsx
+++ b/src/patients/appointments/AppointmentsList.tsx
@@ -28,11 +28,7 @@ const AppointmentsList = (props: Props) => {
     <ul style={{ whiteSpace: 'pre-line' }}>
       {appointments.map((a) => (
         <ListItem action key={a.id} onClick={() => history.push(`/appointments/${a.id}`)}>
-          {`${t('scheduling.appointment.location')}: ${a.location}
-          ${t('scheduling.appointment.reason')}: ${a.reason}
-          ${t('scheduling.appointment.type')}: ${a.type}
-          ${t('scheduling.appointment.startDate')}: ${new Date(a.startDateTime).toLocaleString()}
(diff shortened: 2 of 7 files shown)


igeagonz:
Unit tests are pending. But wanted to get this out to start getting it reviewed.

src/patients/appointments/AppointmentsList.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }, [dispatch, patientId])
+
+  const list = (
+    // inline style added to pick up on newlines for string literal
+    <ul style={{ whiteSpace: 'pre-line' }}>
+      {appointments.map((a) => (
+        <ListItem action key={a.id} onClick={() => history.push(`/appointments/${a.id}`)}>
+          {`${t('scheduling.appointment.location')}: ${a.location}
+          ${t('scheduling.appointment.reason')}: ${a.reason}
+          ${t('scheduling.appointment.type')}: ${a.type}
```
igeagonz: location, reason and type aren't required fields when creating an appointment. Could conditionally render, but this would get ugly... unless someone has an idea 😄

src/patients/appointments/AppointmentsList.tsx
```diff
@@ -0,0 +1,75 @@
...
+          ${t('scheduling.appointment.startDate')}: ${new Date(a.startDateTime).toLocaleString()}
+          ${t('scheduling.appointment.endDate')}: ${new Date(a.endDateTime).toLocaleString()}`}
```
jackcmeyer: I think for now, let's simply display the start date time of appointment. 

Something like `2020-01-02 10:00am` (except the internationalized version of that string).

Eventually, once the `Table` component is finished, we can migrate to that.

src/patients/appointments/AppointmentsList.tsx
```diff
@@ -0,0 +1,75 @@
...
+        <div className="input-group" style={{ width: '100%' }}>
+          <TextInput
+            size="lg"
+            value={searchText}
+            placeholder={t('actions.search')}
+            onChange={onSearchBoxChange}
+          />
+          <div className="input-group-append">
+            <Button onClick={onSearchFormSubmit}>{t('actions.search')}</Button>
+          </div>
```
jackcmeyer: Now that we have two search boxes, we should consider abstracting that to a component in the `/components` folder.

src/clients/db/AppointmentsRepository.ts
```diff
@@ -6,6 +6,38 @@ export class AppointmentRepository extends Repository<Appointment> {
   constructor() {
     super(appointments)
   }
+
+  // Fuzzy search for patient appointments. Used for patient appointment search bar
+  async searchPatientAppointments(patientId: string, text: string): Promise<Appointment[]> {
```
jackcmeyer: Searching this list is not something that I had previously considered. I think that this is a really neat idea!

@fox1t @tehKapa @lauggh @StefanoMiC, 

What does everyone thing the requirements should be here? I see a really cool opportunity for having filter buttons for the different appointment types.

jackcmeyer:
@igeagonz thanks for your contribution 🎉

src/patients/appointments/AppointmentsList.tsx
igeagonz: Ok. I'll simply reduce it to just the Date.toLocaleString() which defaults to the locale it's running on.

src/clients/db/AppointmentsRepository.ts
fox1t: We can leave it like this for now. We need to open a new feature issue to search about this. We really want a multi-field fuzzy search on almost every entity.

fox1t:
Thanks for your contribution! 🎉🎉🎉🎉🎉

PR #1868, opened 2020-02-27
fix(breadcrumb): set appointment date precision to minutes in breadcrumb

### LINKED ISSUE #1853: Breadcrumbs shows time to seconds precision when viewing an appointment
## 🐛 Bug Report

When viewing an appointment, the breadcrumb link shows the start date and end date the seconds precision. 

## To Reproduce

1. Create a new appointment
2. Click on the appointment in the calendar
3. See the breadcrumb

image

## Expected behavior

The time should only be precise to the minutes since the seconds are not relevant for appointment start date and end date.

oliv37:
fix #1853

Set appointment date label precision to minutes


fox1t:
LGTM

PR #1875, opened 2020-03-03
feat(patients): new related person alert

### LINKED ISSUE #1764: Successful alert when new related person is added
## Scenario
**AS**

A user with write access to a patient

**I WANT**

A notification that I have successfully added a related person to a patient

**SO THAT**

I am not confused to whether or not the related person has been successfully added.

## Acceptance Criteria
**GIVEN**

Write access to the patient

AND

I have clicked the add new related person button after successfully filing out the new related person form

**WHEN**

The related person is successfully created

**THEN**

I should receive a toast in the top left corner saying I have successfully added the new related person

oizuldan:
Fixes #1764.

- Added a success alert on top-left corner when a new related person is added to a patient
- changed react-scripts version to tackle the problem https://app.slack.com/client/T02BT7ZBB/C02BT7ZBK


src/patients/patient-slice.ts
```diff
@@ -71,7 +71,11 @@ export const createPatient = (patient: Patient, history: any): AppThunk => async
   )
 }
 
-export const updatePatient = (patient: Patient, history: any): AppThunk => async (dispatch) => {
+export const updatePatient = (
```
jackcmeyer: I think instead of having a flag for for `relatedPersons` we should dispatch a different action. 

I think creating a function like the below gives us more flexibility and better traceability into what actions have been executed..
```ts
export const addRelatedPerson = (patient: Patient, history: any): AppThunk = async (dispatch) => {
 ...
}
```

We can then create a helper function to do any common actions between the update code and the add related persons code.

src/patients/related-persons/RelatedPersonTab.tsx
```diff
@@ -80,7 +80,7 @@ const RelatedPersonTab = (props: Props) => {
       relatedPersons: newRelatedPersons,
     }
 
-    dispatch(updatePatient(patientToUpdate, history))
+    dispatch(updatePatient(patientToUpdate, history, true))
```
jackcmeyer: Using the recommended new action approach, this code would become

```ts
dispatch(addRelatedPerson(patientToUpdate, history))
```

src/patients/patient-slice.ts
oizuldan: added

PR #1880, opened 2020-03-04
Model updates

### LINKED ISSUE #1876: Use uniquely generated code instead of progressive number
## 🐛 Bug Report

Since CouchDB is a really distributed database, we need to be sure that unique fields are really unique, even if two distinct users adds same resource at the same time.
Starting from this postulate, the method https://github.com/HospitalRun/hospitalrun-frontend/blob/master/src/clients/db/PatientRepository.ts#L7 doesn't respects it. This is a recommandation coming both, from PouchDB and CouchDB docs.

## To Reproduce

Two users add a new user at the same time, and PouchDB/CouchDB sync is fired at the same time.

## Expected behavior

* Patient has to have human readable unique identifier to use instead of the _id. We might use https://www.npmjs.com/package/shortid to generate it. 
* Rename `friendlyId` to `code`

### LINKED ISSUE #1877: Use UUID instead of dateTime for PouchDB/CouchDB _id property
## 🐛 Bug Report
Use [UUID version 4](https://www.npmjs.com/package/uuid#version-4-random) to generate cryptographically-strong random values and don't leak any further information. Store it in _id prop of every object that is going to be saved on the DBs.

### LINKED ISSUE #1879: Store created date and last updated date as part of a CouchDB model when saving/updating
## 🚀 Feature Proposal

For auditing purposes, we would like to save the created date and last updated date for CouchDB models. 

These should be stored as unix timestamps.

Created date should only be updated on object create. 
Last Updated should be displayed anytime the model is updated.

## Motivation

Auditing

## Example
```ts
{
	// .... other fields in patient
	createdDate: ${unix timestamp},
	lastUpdatedDate: ${unixTimestamp}, 
}
```

jackcmeyer:
Fixes #1876 
Fixes #1877
Fixes #1879 

- change patient friendly id to be patient code and change it from sequentially generating number to randomly generated string
- change database model ids from timestamp based ids to randomly generated uuids
- adds created date and last updated date to database models

- [uuid](https://bundlephobia.com/result?p=uuid@7.0.1)
- [shortid](https://bundlephobia.com/result?p=shortid@2.2.15)


src/clients/db/Repository.ts
```diff
@@ -47,8 +48,15 @@ export default class Repository<T extends AbstractDBModel> {
   }
 
   async save(entity: T): Promise<T> {
+    const currentTime = getTime(new Date())
```
fox1t: I really prefer saving dates in `ISOString`. It has same properties of ordering like unix timestamp but it is more readable. Can you change it to [toISOString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) everywhere?

src/clients/db/Repository.ts
```diff
@@ -47,8 +48,15 @@ export default class Repository<T extends AbstractDBModel> {
...
+
     const { id, rev, ...valuesToSave } = entity
-    const savedEntity = await this.db.put({ _id: getTime(new Date()).toString(), ...valuesToSave })
+    const savedEntity = await this.db.put({
+      _id: uuidv4(),
+      ...valuesToSave,
+      createdDate: currentTime,
```
fox1t: I think we can use what is the standard defacto for naming these fields:
- `createdDate` -> `createdAt`
- `lastUpdatedDate` ->`updatedAt`

And use ISOString as format, as I already pointed out.

fox1t:
LGTM! Nicely done.

PR #1897, opened 2020-03-12
feat: added successful alert when adding new appointments

### LINKED ISSUE #1894: Success alert when new appointment is created
## 🚀 Feature Proposal

When a new appointment is created, the user should be notified with a success message.

## Motivation

The user should be alerted when they do something.

sotous:
Added successful alert when adding new appointments. Replicated successful alert from new patients process and added the successful property on the translation file.

Fixes #1894 

- Added a property to the `index.ts` which belongs to `enUs` translations for `scheduling`, in order to display correctly the successful message.
- Modified `NewAppointment.tsx` file. The function `onNewAppointmentSaveSuccess`, was modified by adding a `toast` to it after the `history.push` process ends successfully.

**There were not new dependencies added.


src/scheduling/appointments/new/NewAppointment.tsx
```diff
@@ -42,6 +42,11 @@ const NewAppointment = () => {
 
   const onNewAppointmentSaveSuccess = (newAppointment: Appointment) => {
     history.push(`/appointments/${newAppointment.id}`)
+    Toast(
+      'success',
+      t('Success!'),
+      `${t('scheduling.appointment.successfullyCreated')} ${newAppointment.id}`,
+    )
```
jackcmeyer: Could we add a test to make sure that the toast is getting called and has the proper message?

PR #1922, opened 2020-03-19
style: makes unused vars error with linting

jackcmeyer:

- makes unused variables error with linting


src/clients/db/Repository.ts
```diff
@@ -1,4 +1,4 @@
-/* eslint "@typescript-eslint/camelcase": "off" */
+/* eslint "@typescript-eslint/camelcase": "off", "@typescript-eslint/no-unused-vars": "off" */
```
jackcmeyer: In general we want to not have unused variables in our code, however, in this file we are restructuring generic objects and ensuring we don't copy over some ids, which is why I have ignored it in this file and this file only.

tehKapa:
@Jack, I found this reference https://stackoverflow.com/questions/57802057/eslint-configuring-no-unused-vars-for-typescript 
we're missing an extension https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin#recommended-configs

src/clients/db/Repository.ts
fox1t: Add this rule https://eslint.org/docs/rules/no-unused-vars#ignorerestsiblings to eslintrc instead.
` { "ignoreRestSiblings": true }`
On object destructuring is always safe to assume that we just want to get rid of some props.

.eslintrc.js
```diff
@@ -7,7 +7,9 @@ module.exports = {
   },
   extends: [
     'airbnb',
-    'plugin:@typescript-eslint/recommended',
+    "eslint:recommended",
+    "plugin:@typescript-eslint/eslint-recommended",
```
fox1t: What's the difference between typescript-eslint/eslint-recommended and typescript-eslint/recommended? I really can't get it.

PR #1944, opened 2020-03-24
feat(patients): added an asterisk to the Allergy Name field

### LINKED ISSUE #1928: New Allergy Modal should mark required fields
## 🚀 Feature Proposal
The name field should be marked with a `*` to indicate that it is required.

## Motivation
Better use experience.

mamoses:
Fixes #1928

- Added an asterisk to indicate that the name field is required


.env.example
```diff
@@ -1 +1 @@
-REACT_APP_HOSPITALRUN_API=http://0.0.0.0:3001
\ No newline at end of file
+REACT_APP_HOSPITALRUN_API=http://0.0.0.0:5984
```
jackcmeyer: Any reason to change the port? 5984 is usually what pouchdb runs on. We want to point to our server (which proxies to pouchdb).

mamoses: In the connecting to the server section I misunderstood step 4. Change the REACT_APP_HOSPITALRUN_API variable to point to the HospitalRun Development Server and had trouble running it. I got it working but forgot to switch it back sorry about that.

jackcmeyer: No worries :)could you switch it back? Then we can merge this PR!

mamoses: I changed it back, let me know if there is any other changes that need to be made! Thanks!

Issue #1976, opened 2020-04-07
Create/Edit Patient inputs are not validated


martimfj:
When creating and editing a patient, only the required `givenName` is validated. The email and tel inputs can be anything and the date of birth can be a day greater than today.

## 🐛 Bug Report

Input fields from Editing/Creating Patients are not validated.

## To Reproduce

Input a text in the telephone field
Input a text that is not an email in the email field
Input a date in 2021 in the date of birth field

The components already give a way to add validation feedback to the user, but it is not used at the time.

BrunaKimura:
Should text fields also be validated to avoid numeric inputs?

For example:
- Suffix
- Preffix
- Given Name
- Family Name
- Occupation
- Preferred Language

LucasBN:
Would like to take a crack at this, could someone assign me?

jackcmeyer:
@LucasBN I think validating addresses may be difficult due to needing to support any address format for any location in the world; however I have not looked into it. I think we can just tack it on to this one.

jackcmeyer:
> Should text fields also be validated to avoid numeric inputs?
> 
> For example:
> 
> * Suffix
> * Preffix
> * Given Name
> * Family Name
> * Occupation
> * Preferred Language

@BrunaKimura I think that preventing numeric inputs is reasonable except for Occupation.

LucasBN:
> @LucasBN I think validating addresses may be difficult due to needing to support any address format for any location in the world; however I have not looked into it. I think we can just tack it on to this one.

Yeah I also thought that might be a problem. I'm going to look into it anyway as I'd like to contribute somewhat - will let you know if I get anywhere. (I'll look at avoiding numeric inputs whilst I'm at it).

jackcmeyer:
> > @LucasBN I think validating addresses may be difficult due to needing to support any address format for any location in the world; however I have not looked into it. I think we can just tack it on to this one.
> 
> Yeah I also thought that might be a problem. I'm going to look into it anyway as I'd like to contribute somewhat - will let you know if I get anywhere. (I'll look at avoiding numeric inputs whilst I'm at it).

sounds great!

PR #1985, opened 2020-04-14
feat(patients): max DOB set for current date

### LINKED ISSUE #1984: Date of Birth date picker should have max date of the current date
## 🚀 Feature Proposal

The Date Picker component supports defining a maximum date. The date of birth date picker should set the maximum date as today. 

## Motivation

Ease of use

jlato89:
Added maxDate prop to DatePicker component to avoid DOB being set in the future.

Fixes #1984.

- Added maxDate prop to DatePickerWithLabelFormGroup component to restrict DOB from being set in the future.

- No new dependencies

### DIFF
--- a/src/components/input/DatePickerWithLabelFormGroup.tsx
+++ b/src/components/input/DatePickerWithLabelFormGroup.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import { Label, DateTimePicker } from '@hospitalrun/components'
+import { addDays } from 'date-fns'
 
 interface Props {
   name: string
@@ -22,6 +23,7 @@ const DatePickerWithLabelFormGroup = (props: Props) => {
         dateFormat="MM/dd/yyyy"
         dateFormatCalendar="LLLL yyyy"
         dropdownMode="scroll"
+        maxDate={addDays(new Date(), 0)}
         selected={value}
         timeIntervals={30}
         withPortal={false}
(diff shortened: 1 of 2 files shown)


jlato89:
This also Fixes #1976 regarding the date of birth

src/components/input/DatePickerWithLabelFormGroup.tsx
```diff
@@ -22,6 +23,7 @@ const DatePickerWithLabelFormGroup = (props: Props) => {
         dateFormat="MM/dd/yyyy"
         dateFormatCalendar="LLLL yyyy"
         dropdownMode="scroll"
+        maxDate={addDays(new Date(), 0)}
```
jackcmeyer, line 26: This component is intended to be more general than for use inside of the new/edit patient screen. 

I think that this interface will only need to expose an optional maxDate and just pass it along to the `DateTimePicker from `@hopsitalrun/components`. 

Then we'll need to make a change in the `GenrealInformationComponent` to set the `maxDate` property.

fox1t: I agree 100% with your vision. @jlato89 can you handle it?

jlato89:
I had a feeling it was but couldn't find another instance of it being used. 
Will work on those changes. Tests might take a bit longer as I havnt done 
them before but I'm sure il figure it out.

src/components/input/DatePickerWithLabelFormGroup.tsx
jlato89: I should be able to yes. just an oversight on my part.

jlato89:
Added the appropriate changes. Just have to figure out how todo the tests. Then it will be ready to be merged.

jackcmeyer:
closing in favor of #2002

PR #1987, opened 2020-04-14
feat(viewpatient): added labs tab to ViewPatient

### LINKED ISSUE #1972: Create labs tab in view patient screen
## 🚀 Feature Proposal
If I have the ability to view labs and view patients, there should be a labs tab in the View Patient screen. 

The labs tab should show a list of labs that are associated with the patient. The table should show the lab type, the lab status, and the requested on date.. 

Clicking on one of the labs should bring me to `/labs/:id`. 

## Motivation
Linking patients and labs.

rubencgt:
A patient can see all his requests for labs in the Labs tab

fix #1972


tehKapa:
Now's build fails are caused by internal incident.

src/clients/db/LabRepository.ts
```diff
@@ -6,6 +6,18 @@ export class LabRepository extends Repository<Lab> {
   constructor() {
     super(labs)
   }
+
+  async searchPatientLabs(patientId: string): Promise<Lab[]> {
```
jackcmeyer, line 10: Let's rename this to be `findLabsByPatientId`

src/patients/labs/Labs.tsx
```diff
@@ -0,0 +1,63 @@
+import React, { useEffect } from 'react'
```
jackcmeyer, line 1: Need tests for this component: 

- [ ] test that the table renders with the correct labs
- [ ] test that the content of the table is correct
- [ ] test that when no labs are returned, the warning message appears

src/labs/labs-slice.ts
```diff
@@ -0,0 +1,42 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
+import Lab from '../model/Lab'
+import { AppThunk } from '../store'
+import LabRepository from '../clients/db/LabRepository'
+
+interface LabsState {
+  isLoading: boolean
+  labs: Lab[]
```
jackcmeyer, line 8: I think that this should be added as part of the patient slice. It seems like this functionality is part of the patient module, so therefore it makes sense to have the patient slice own this state update. @fox1t any thoughts here?

I imagine the patient slice getting a new state field called `labs`, and then a thunk function called `findLabsPatientId` and the corresponding reducers.

src/labs/labs-slice.ts
```diff
@@ -0,0 +1,42 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
```
jackcmeyer, line 1: Need to add tests for this functionality. Of course the test implementation might change a little if we move it to the patient slice, however, the test concepts shouldn't. 

- [ ] test for the `fetchPatientLabs` thunk dispatching `fetchLabsStart`
- [ ] test that it calls the `LabRepository` correctly
- [ ] test that it dispatches `fetchLabsSuccesss` with the correct data
- [ ] test that the `fetchLabsStart` reducer update updates the is loading state correctly
- [ ] test that the `fetchLabsSuccess` reducer updates the labs list and loading state correctly

src/clients/db/LabRepository.ts
jackcmeyer, line 10: Need to add tests to show that this is working correctly.

src/labs/labs-slice.ts
rubencgt: in that case, the Labs should be stored using the src/patients/patient-slice.ts:73?
instead of LabsRepository?

rubencgt:
@jackcmeyer Hi Jack,
I've applied the changes requested, 
please let me know if it is ok, or it needs more work

src/model/Patient.ts
```diff
@@ -18,4 +19,5 @@ export default interface Patient extends AbstractDBModel, Name, ContactInformati
   allergies?: Allergy[]
   diagnoses?: Diagnosis[]
   notes?: Note[]
+  labs?: Lab[]
```
jackcmeyer, line 22: I don't think that labs should be a property of a Patient at the database level. A Patient doesn't really "own" a lab a patient is just a property of the lab, which is why I think it makes sense to "link" the patient to the lab in the Lab entity. This will also help with security models going forward since someone who might not be able to see patient information can see lab information.

@fox1t thoughts here?

rubencgt: you are right,
then I propose remove this because the link already exists since the `Lab.ts` contains already the patient id at the database level.

and in Patient Lab Tab I would use LabRepository to find all the labs related to this patient, and then render them in the Lab Tab.

does that sound good?

jackcmeyer: sounds good to me!

rubencgt: I have done, the discussed changes,
please let me know if it is ok

jackcmeyer: @rubencgt code changes look good! there is just a failing test to take care of. I think there may be a missing mock somewhere. See the failing check for the logs.

jackcmeyer:
@rubencgt the MacOS build is failing. I'm guessing due to a missing mock somewhere. Let me know if you need help resolving it.

rubencgt:
@jackcmeyer for some reason I can't reproduce the issue locally (I use MacOS 10.15.4), 
but I added the mock in the test where it seemed to be it was missing.

jackcmeyer:
> @jackcmeyer for some reason I can't reproduce the issue locally (I use MacOS 10.15.4),
> but I added the mock in the test where it seemed to be it was missing, NewAppointment.test.tsx.

I was able to reproduce (I also use a Mac). I'll investigate a fix.

PR #2003, opened 2020-04-21
feat(appointments) reason field no longer labelled as a required field when making appointments.

### LINKED ISSUE #1997: Reason is incorrectly being marked as required field on edit/new appointment screen
## 🐛 Bug Report
The reason field is being marked as required when it should not be. 

## To Reproduce
1. Navigate to edit or create appointment screen 

## Expected behavior
The reason should not be marked as required.

JDarke:
Fixes #1997 .

changed line 21 of src/components/input/textfieldwithlabelformgroup.tsx , so that Label attr 'isRequired' now takes value from Props.isRequired? (line 11).


package.json
```diff
@@ -31,7 +31,8 @@
     "redux-thunk": "~2.3.0",
     "shortid": "^2.2.15",
     "typescript": "~3.8.2",
-    "uuid": "^7.0.1"
+    "uuid": "^7.0.1",
+    "yarn": "~1.22.4"
```
jackcmeyer, line 35: You are right, this can be removed. Yarn is expected to be globally installed.

src/components/input/TextFieldWithLabelFormGroup.tsx
```diff
@@ -14,11 +14,11 @@ interface Props {
 }
 
 const TextFieldWithLabelFormGroup = (props: Props) => {
-  const { value, label, name, isEditable, isInvalid, feedback, onChange } = props
+  const { value, label, name, isEditable, isInvalid, isRequired, feedback, onChange } = props
   const id = `${name}TextField`
   return (
     <div className="form-group">
-      <Label text={label} htmlFor={id} isRequired />
+      <Label text={label} htmlFor={id} isRequired={isRequired} />
```
jackcmeyer, line 21: can we add a test that, if `isRequired` is provided`, that it is passed to the label and marks the label as required

JDarke: As far as I can see, that's tested when FormLabel is imported into Label.tsx...see the attached screenshot (it seemed more useful to use a pic in this instance than to just post the code - apologies if that's incorrect practice!)

The /src/components path contains a lot more subfolders than my local clone, so I can't access a local copy of Label.tsx.  Is that being compiled on the server side?  

Screen Shot 2020-04-21 at 18 15 49

package.json
JDarke: Ok, I'll take that out in the next commit.

src/components/input/TextFieldWithLabelFormGroup.tsx
jackcmeyer: The components that you are seeing are coming from this library: 
https://github.com/HospitalRun/components

For this test, all we want to verify is that if the `isRequired` prop is given to the `TextFieldWithLabelFormGroup` component, that is passes it to the `Label` component. 

It doesn't actually verify that the functionality is working, per se. We trust that the implementation of the `Label` component works as intended. What we are testing is that we are using the `Label` component correctly. See this test for how we tested the that the correct text is used: 
https://github.com/HospitalRun/hospitalrun-frontend/blob/c703928c3e2ded25207d7f2ecfbf7a0441743ef8/src/__tests__/components/input/TextFieldWithLabelFormGroup.test.tsx#L7

JDarke: Ah, I see...yup, I'll give it a go.

JDarke: Is the test to be added into TextFieldWithLabelFormGroup.test.tsx (with the text field test suite you referenced) or should it be a separate local test in TextFieldWithLabelFormGroup.tsx?

jackcmeyer: Either! I don't have a preference

JDarke: I've got a new commit to push, but it's telling me that my local and remote have diverged.  Tried a pull to merge them, and my terminal locked up.  Restarted and retired, and now I have this message in terminal:

```
error: You have not concluded your merge (MERGE_HEAD exists).
hint: Please, commit your changes before merging.
fatal: Exiting because of unfinished merge.
```

Is this fixable, or do I need to save my local changes, re-clone the repo to local, and then send a whole new PR, or does it need to be connected to this thread's PR?

JDarke: Here's the test, btw:

```
describe('text field with label form group', () => {
  describe('layout', () => {
    it('should render a label', () => {
      const expectedName = 'test'
      const wrapper = shallow(
        <TextFieldWithLabelFormGroup
          name={expectedName}
          label="test"
          value=""
          isEditable
          onChange={jest.fn()}
        />,
      )

      const label = wrapper.find(Label)
      expect(label).toHaveLength(1)
      expect(label.prop('htmlFor')).toEqual(`${expectedName}TextField`)
      expect(label.prop('text')).toEqual(expectedName)
    })

    it('should render label as required if isRequired is true', () => {
      const expectedName = 'test'
      const wrapper = shallow(
        <TextFieldWithLabelFormGroup
          name={expectedName}
          label="test"
          value=""
          isEditable
          isRequired={true}
          onChange={jest.fn()}
        />,
      )

      const label = wrapper.find(Label)
      expect(label).toHaveLength(1)
      expect(label.prop('isRequired')).toBeTruthy()
    })

```

JDarke: Ah, I seem to have got it pushed now, and it looks like the earlier ones did, too.  Apologies if I'm cluttering up this PR with multiple commits.

Issue #1966, opened 2020-04-05
Labs should sort by requested date by default


jackcmeyer:
## 🚀 Feature Proposal
Labs should sort by requested date by default (the newest one should come first) when I am viewing the list of labs

## Motivation
Easier to find lab requests

## Blocked by
#1969

akshay-ap:
I would like to work on this.

jackcmeyer:
@akshay-ap sounds great! I've assigned this to you.

akshay-ap:
@jackcmeyer Before starting to work on it, I have a suggestion: A new dropdown which will show the fields by which the list of Labs can be sorted in ascending or descending order: 
1. ResquestedOn 
2. Type 
3. Status  
By default, list will be sorted by ResquestedOn descending order.

![image](https://user-images.githubusercontent.com/16358982/79756052-a8ffd580-8337-11ea-97af-eb85044c983d.png)

jackcmeyer:
@akshay-ap

 I like this idea. However, most of the sorting functionality is going to be fully implemented inside of the Table component. 

For now, the intent of this was to get basic functionality implemented so that it is a usable feature. 

If the solution that you are suggesting is simple enough, then I'd say go for it. However, it is likely that the UI part of the story will change in the future.

jackcmeyer:
Closed via #2004

PR #2004, opened 2020-04-21
feat(viewlabs): sort Labs by requestedOn field

akshay-ap:
feat #1966

- Sort the list of Labs by requestedOn field


src/labs/ViewLabs.tsx
```diff
@@ -38,7 +39,15 @@ const ViewLabs = () => {
 
   useEffect(() => {
     const fetch = async () => {
-      const fetchedLabs = await LabRepository.findAll()
+      const sortRequest: SortRequest = {
```
jackcmeyer, line 42: A test should be added to cover this scenario. 

In this case, I think it is sufficient to test that the `findAll` method was called with the correct sort request.

akshay-ap: I have added the test.  Please let me know if any change is needed.

PR #2012, opened 2020-04-24
fix(patients): fixes search when using special characters

### LINKED ISSUE #1999: Error while searching Patient
## 🐛 Bug Report
```
Uncaught (in promise) SyntaxError: Invalid regular expression: /Adk\/: \ at end of pattern
    at RegExp (<anonymous>)
    at a.<anonymous> (PatientRepository.ts:21)
    at s (2.726afb2a.chunk.js?__WB_REVISION__=15200f807ae5907c400b:sourcemap:2)
    at Generator._invoke (2.726afb2a.chunk.js?__WB_REVISION__=15200f807ae5907c400b:sourcemap:2)
    at Generator.next (2.726afb2a.chunk.js?__WB_REVISION__=15200f807ae5907c400b:sourcemap:2)
    at r (2.726afb2a.chunk.js?__WB_REVISION__=15200f807ae5907c400b:sourcemap:2)
    at c (2.726afb2a.chunk.js?__WB_REVISION__=15200f807ae5907c400b:sourcemap:2)
    at 2.726afb2a.chunk.js?__WB_REVISION__=15200f807ae5907c400b:sourcemap:2
    at new Promise (<anonymous>)
    at a.<anonymous> (2.726afb2a.chunk.js?__WB_REVISION__=15200f807ae5907c400b:sourcemap:2)
```

## To Reproduce

1. Create a patient
2. Enter text like 'Adk/\'(string containing '/\') in patient search bar.

## Expected behavior

No error.

## Your Environment

- *node version*: 12
- *fastify version*: >=1.0.0
- *os*: Windows

JDarke:
Fixes #1999.

Search string in patient search field is now cleaned of any breaking special characters.

Removes any number of occurrences of these chars:  
\[*?()

<a href="https://gitpod.io/#https://github.com/HospitalRun/hospitalrun-frontend/pull/2012"></a>


src/clients/db/PatientRepository.ts
```diff
@@ -13,12 +13,13 @@ export class PatientRepository extends Repository<Patient> {
   }
 
   async search(text: string): Promise<Patient[]> {
+    const cleanText = text.replace(/[\\[()?*]+/g, '')
```
akshay-ap, line 16: I also checked with String ```{1}```.
```SyntaxError: Invalid regular expression: /{1}/: Nothing to repeat.```
Need to handle this case as well.

Comment (jackcmeyer) [reply src/clients/db/PatientRepository.ts:16]: I think that we should also remove `]`, and `{`, and `}`

Comment (JDarke) [reply src/clients/db/PatientRepository.ts:16]: Ok, will do.

**Comment by JDarke** [comment]
Sorry, I sent a repeated commit;  my initial commit push failed due to missing some refs that were now in the origin, so used git pull and then re-committed and pushed....looks like the first one went through anyway.


Ok, so I've written a test into the search suite in src/__tests__/clients/db/PatientRepository.test.ts, which I think should do the trick...but I haven't yet worked out how to run or view results from the tests.

File: src/clients/db/PatientRepository.ts (hunk 1 above)
Comment (fox1t) [reply src/clients/db/PatientRepository.ts]: I think that we need to remove all valid regex chars. Is there any lib out there that can help us to not forget any char?

File: package.json [hunk 2]
```diff
@@ -13,6 +13,7 @@
     "i18next": "~19.4.0",
     "i18next-browser-languagedetector": "~4.1.0",
     "i18next-xhr-backend": "~3.2.2",
+    "mocha": "~7.1.1",
```
fox1t, line 16: Why adding mocha? We already use Create React App integrated test runner (Jest).

src/clients/db/PatientRepository.ts
```diff
@@ -13,12 +13,13 @@ export class PatientRepository extends Repository<Patient> {
   }
 
   async search(text: string): Promise<Patient[]> {
+    const cleanText = text.replace(/[\\[\](){}?*]+/g, '')
```
fox1t, line 16: Maybe is better to use some stable package to search for char to leave out from a regexp? What do you think?

package.json
JDarke: Aha, I see.   I got confused a bit mixed up reading up on testing nodejs.  Thanks for that...I'll take out the mocha dependency in my next commit.

src/clients/db/PatientRepository.ts
JDarke: Not sure what you mean by stable package...?

Do you mean something instead of the replace(regexp,'') method?

JDarke:
I'm not sure what's happened here...my previous pushes passed the checks;  what have I messed up...?

src/clients/db/PatientRepository.ts
fox1t: You can use this https://www.npmjs.com/package/escape-string-regexp
It is ok to add a new dependency in this case.

JDarke: That particular package also removes hyphens, so we'd need to modify it.  

I'm new to TypeScript & node;  what would be the benefit of using a package in this instance?

jackcmeyer:
> I'm not sure what's happened here...my previous pushes passed the checks; what have I messed up...?

I think it was just a glitch in the matrix. Everything looks back to normal.

src/clients/db/PatientRepository.ts
fox1t: The benefit is that we are sure that we never miss any possible value. Since we will need this logic also on the server side and maybe in core package, it is better to use external package instead of copy-pasting the same code over several repositories.

JDarke: Ok.  What's the procedure for modifying existing packages?

Do I need to clone and modify the package you suggested (as we need to leave hyphens intact) and then install it, or do I add it to the master and edit it from within the project?

fox1t: It is easier like that. Just go in the root of this project and user yarn to add it as dependency
`yarn add escape-string-regexp` then go in this file and import it like `import escapeRegexp from 'escape-string-regexp'` and then you can use it in place of `.replace()`.

fox1t: For the hypens, they will be escaped, and that's what we need. Am I missing something?

JDarke: Hyphens are a relatively common character, especially in Western names.  Double-barrelled surnames names like John Cooper-Clarke are common in English, and hyphenated first names are common in French, for example;  'Jean-Pierre', or 'Anne-Marie' is not first-and-middle-name, it's just their first name.

JDarke: Hyphens are a relatively common character, especially in Western names.  Double-barrelled surnames names like John Cooper-Clarke are common in English, and hyphenated first names are common in French, for example;  'Jean-Pierre', or 'Anne-Marie' is not first-and-middle-name...it's just their first name.

fox1t: What i meant is that we just need to escape that chars, not to remove them. That's why I am suggesting to use that package.

JDarke: Ohhhhh.  *facepalm* . 

Yeah, I understand what you mean now.  Sorry, I was conflating escape with removal for some reason.  

Yes, that's a much neater approach, thank you.  I'll give it a go!

fox1t: No worries! I appreciate when there is a dialog on PR, since we it is important to not missing anything!
And, of course, thanks for your time and contribution!

JDarke: Ok, so I installed the package (though I still can't seem to get yarn to work, so used npm), and the new code seems to work fine:

```
async search(text: string): Promise<Patient[]> {
    const escapedString = escapeStringRegexp(text)
    return super.search({
      selector: {
        $or: [
          {
            fullName: {
              $regex: RegExp(escapedString, 'i'),
            },
          },
          {
            code: text,
          },
        ],
      },
    })
  }
```

However, the test suite that I wrote for the previous string cleaning method also needed to be updated, and it's failing the test:

```
describe('search', () => {
    afterEach(async () => {
      await removeAllDocs()
    })

    it('should escape all special chars from search text', async () => {
      await patients.put({ _id: 'id9999', code: 'P00001', fullName: 'test -]?}(){*[\\$+.^test' })

      const result = await PatientRepository.search(
        'test \\-\\]\\?\\}\\(\\)\\{\\*\\[\\\\\\$\\+\\.\\^test',
      )

      expect(result).toHaveLength(1)
      expect(result[0].id).toEqual('id9999')
    })

```

It's coming back as result length = 0, so it's not finding a match.

Am I misunderstanding what the result string should be? 

Or perhaps something to do with the method of the new package inserting a \unnnn escape:

> Use a simple backslash escape when it’s always valid, and a \unnnn escape when the simpler form would be disallowed by Unicode patterns’ stricter grammar.

As I understand it, the escaped string should be the same as the input string, but with the special chars escaped with a backslash...or am I trying to read backslashes that aren't there, because they were ignored?

JDarke: Am I perhaps overcomplicating this, and only need to test that the input and output strings match directly?

Issue #1967, opened 2020-04-05
Patient list should sort by given name ascending by deafult


jackcmeyer:
## 🚀 Feature Proposal
By default, patients should be sorted by given name ascending. 

## Motivation
The list should not be sorted in a random way.

sujithnath:
@jackcmeyer  Can I take it?

vicdurai:
Can I pick this issue or is someone working on this already?

akshay-ap:
FYI: The patient list will be sorted by name in #2026.

vicdurai:
Cool since this is being worked on, I will not work on this. @tehKapa Please can you unassign me from this issue and assign it to @akshay-ap since he already has a PR for it.

tehKapa:
Ok. Done! Sorry @akshay-ap

PR #2026, opened 2020-04-28
feat(patient): add paging and sorting functionality

### LINKED ISSUE #1969: Add paging and sorting functionality to find all/search functionality
## 🚀 Feature Proposal

Add paging and sorting functionality to the search/find all functions for patients, appointments, lab.

## Motivation
* Patient Search/List, Appointment List, Lab List

akshay-ap:
Fixes #1969

- New component `PageComponent` (Shows: Previous Button, Current Page Number, Next Button)
- PageRequest Interface
- Wrap `Patient[]` in `Page<Patient>`
- Add index on `Patient`
- Add methods `searchPaged` and `findAllPaged` in PatientRepository
- Changes in test cases


akshay-ap:
- Added methods `searchPaged` and `findAllPaged` in PatientRepository to avoid any breaking changes in other Components.
Once changes looks good, `searchPaged` can be replaced with `serach` and `findAllPaged` with `findAllPaged` respectively and paging can be extended to `Labs`.

src/clients/db/Repository.ts
```diff
@@ -41,6 +43,35 @@ export default class Repository<T extends AbstractDBModel> {
...
+
+    sort.sorts.forEach((s) => {
+      selector[s.field] = { $gt: null }
+    })
+
+    const result = await this.db.find({
+      selector,
+      sort: sort.sorts.length > 0 ? sort.sorts.map((s) => ({ [s.field]: s.direction })) : undefined,
+      limit: pageRequest.limit,
+      skip: pageRequest.skip,
```
jackcmeyer, line 59: In PouchDB, skip is very slow. It literally is doing what it says, "skips" the elements that we don't care about. This means that the elements that are getting "skipped" are still loaded in to memory. 

Checkout this blog post: https://pouchdb.com/2014/04/14/pagination-strategies-with-pouchdb.html.

it provides a mechanism for "smart" skip.

jackcmeyer: To accomplish this, our PageRequest may need to know the previous start key. However there could be a variety of implementations.

src/clients/db/PatientRepository.ts
```diff
@@ -31,6 +36,45 @@ export class PatientRepository extends Repository<Patient> {
...
+              hasPrevious: pageRequest.skip > 0,
+            }
+            resolve(pagedResult)
+          }),
+      )
+      .catch((err) => {
+        console.log(err)
+        return err
+      })
+  }
```
jackcmeyer, line 76: prefer this to be `async`/`await` for easier readability.

src/clients/db/PatientRepository.ts
```diff
@@ -31,6 +36,45 @@ export class PatientRepository extends Repository<Patient> {
     })
   }
 
+  async searchPaged(
+    text: string,
+    pageRequest: PageRequest = UnpagedRequest,
```
jackcmeyer, line 41: sorting muse be included in the search method.

src/clients/db/PatientRepository.ts
```diff
@@ -1,8 +1,10 @@
 import escapeStringRegexp from 'escape-string-regexp'
```
jackcmeyer, line 1: We should add tests in the `PatientRepository.test.tsx` to validate that this is working properly.

src/clients/db/Repository.ts
akshay-ap: Agreed that skip is an expensive operation. I intend to add index on `fullName`, run query on the index. https://pouchdb.com/guides/mango-queries.html#pagination.
But, I am facing issue in pagination:
1. `fullname` cannot be unique (thus needs to be coupled with other field which is a sequence)
2. As `code` or `_id` field is not a sequence, the pagination logic will fail.
currently,  assume records : ``[...{fullName: def, code: P-zzz}, {fullName: ghi, code: P-aaa}...]``
and if last record in a page is `{fullName: def, code: P-zzz}` on clicking next page,
query will be: find docs with `fullName > abc`  and `code > P-zzz`.
Query will not return `{fullName: ghi, code: P-aaa}`.
I propose to change `code` to a sequence or add new field that will be sequence(redundant).

akshay-ap: I think index on `[fullName, _id]` will work.

akshay-ap: I tried with `[fullName, _id]` but did not work. The best solution I can think of is add new field `index` which will be `fullName + code` and thus make sure uniqueness along with indexing.

src/clients/db/PatientRepository.ts
akshay-ap: I have added some tests.

tehKapa:
@jackcmeyer is your change request still valid?

src/clients/db/PageRequest.ts
```diff
@@ -0,0 +1,13 @@
+export default interface PageRequest {
+  number: number | undefined
+  size: number | undefined
+  nextPageInfo: { [key: string]: string | null } | undefined
+  previousPageInfo: { [key: string]: string | null } | undefined
+  direction?: 'previous' | 'next'
+}
+export const UnpagedRequest: PageRequest = {
```
jackcmeyer, line 8: real nitpick here: there should be a new line between 7 & 8

src/hooks/useUpdateEffect.ts
```diff
@@ -0,0 +1,14 @@
+import { useRef, useEffect } from 'react'
```
jackcmeyer, line 1: could you explain what this hook is doing. I'm not sure I understand its intent.

src/model/Patient.ts
```diff
@@ -18,4 +18,5 @@ export default interface Patient extends AbstractDBModel, Name, ContactInformati
   allergies?: Allergy[]
   diagnoses?: Diagnosis[]
   notes?: Note[]
+  index: string
```
jackcmeyer, line 21: Could you explain the need for a separate index field other than the id?

src/clients/db/PatientRepository.ts
```diff
@@ -28,11 +34,72 @@ class PatientRepository extends Repository<Patient> {
...
+
+    const pagedResult: Page<Patient> = {
+      content: result.slice(
+        0,
+        pageRequest.size
+          ? result.length < pageRequest.size
+            ? result.length
+            : pageRequest.size
+          : result.length,
+      ),
```
jackcmeyer, line 82: I think that this should be handled by the database. Not sure that we should need to do a slice after we have a search returned.

tehKapa:
@akshay-ap do you need help?

src/model/Patient.ts
akshay-ap: For paging, the field on which indexing is applied should be unique.
1. Paging with `id` field as index will make the records to appear to be randomly sorted by `fullName`.  
2. Assuming that records should be sorted by `fullName` by default, and `fullName` is non-unique, I thought of combining `fullName` with `patient code` so that uniqueness is maintained.

src/hooks/useUpdateEffect.ts
akshay-ap: https://reactjs.org/docs/hooks-reference.html#useref
The only purpose of using `useRef` here is to store the boolean value.

src/clients/db/PatientRepository.ts
akshay-ap: In the search query, one extra record is fetched.

e.g. if pagesize is 25, search query will search for 26 records. This is to know information about the next page.
But, while returning the result, only 25 records should be returned.
This line simply skips last record.

src/model/Patient.ts
akshay-ap: This also fixes #1967

akshay-ap:
> @akshay-ap do you need help?

Any kind of help will be welcome!!! Code review, contribution, suggestions for this PR. :)

src/clients/db/PageRequest.ts
akshay-ap: Done

src/hooks/useUpdateEffect.ts
akshay-ap: I have added `useUpdateEffectHook` so that the hook is called only after any of the dependency changes and skip during initial render.
Use case: In initial loading of patient info, `searchPatients` should be called only once.
Later,`searchPatients` should be called when `userPageRequest` changes.

akshay-ap: I thought of adding this hook because adding just `userPageRequest ` in dependency array  didn't work here https://github.com/HospitalRun/hospitalrun-frontend/blob/6d15f9fe91af0407066bba430e70d770a1a2fe96/src/patients/list/ViewPatients.tsx#L85

So, thought of segregating `debouncedSearchText` and `userPageRequest` on change callbacks.

src/model/Patient.ts
akshay-ap: @jackcmeyer , can you please provide any feedback on this?

PR #2032, opened 2020-05-01
feat(patient): add input validation

### LINKED ISSUE #1976: Create/Edit Patient inputs are not validated

LucasBN:
Fixes #1976

- Prefix / Suffix / Family Name / Preferred Language Validated to ensure no numerical input
- Added to both edit and create
- Provides feedback to user
- Added tests for new code


tehKapa:
@LucasBN can you resolve the 3 conflicts? thanks!

LucasBN:
> @LucasBN can you resolve the 3 conflicts? thanks!

@tehKapa Done

src/__tests__/patients/patient-slice.test.ts
```diff
@@ -203,6 +203,11 @@ describe('patients slice', () => {
       const expectedPatient = {
         id: expectedPatientId,
         givenName: undefined,
+        dateOfBirth: addDays(new Date(), 4).toISOString(),
+        suffix: '061002',
+        prefix: '061002',
+        familyName: '061002',
+        preferredLanguage: '061002',
```
jackcmeyer, line 210: This test was changed to only check for required fields. I think that we should remove everything but `givenName` in this test. 

I think we'll need to add some other test which checks for numbers in the string.

jackcmeyer: I believe this is why the build is failing.

src/__tests__/patients/patient-slice.test.ts
```diff
@@ -386,6 +395,10 @@ describe('patients slice', () => {
         id: expectedPatientId,
         givenName: undefined,
         dateOfBirth: addDays(new Date(), 4).toISOString(),
+        suffix: '061002',
```
jackcmeyer, line 398: let's change this test to only test the numbers in the values.

PR #2034, opened 2020-05-02
feat(i18n): add new script for checking missing translations

### LINKED ISSUE #1919: Warn about missing translations
## 🚀 Feature Proposal

I would like the application to warn about missing translations when starting the application or running tests. 

A preferable solution would be to have a library that does this for us, which could mean switching internationalization libraries. 

However, as a fallback, a custom script that runs as part of `yarn start` or `yarn test` should work as well. 

## Motivation

Translations are very hard to manage and they get out of sync very quickly.

marcosvega91:
Fixes #1919 

- I added new dep for running typscript script(ts-node)
-  I updated the linter path for checking the new
folder scripts
- I added a new command yarn translation:check to check if some translations are
missing

When you run the command It will print some information of the library i18next because the debug is active

- https://bundlephobia.com/result?p=ts-node@8.9.1
- https://bundlephobia.com/result?p=chalk@4.0.0

_Note: pull requests without proper descriptions may simply be closed without further discussion. We appreciate your contributions, but need to know what you are offering in clearly described format. Provide tests for all code that you add/modify. If you add/modify any components update the storybook. Thanks! (you can delete this text)_


marcosvega91:
I didn't understend why webstorm create a file json in the root 🤮. I deleted it

package.json
```diff
@@ -109,11 +109,12 @@
...
     "test:ci": "cross-env CI=true react-scripts test --passWithNoTests",
-    "lint": "eslint \"src/**/*.{js,jsx,ts,tsx}\"",
-    "lint:fix": "eslint \"src/**/*.{js,jsx,ts,tsx}\" --fix",
+    "lint": "eslint \"src/**/*.{js,jsx,ts,tsx}\" \"scripts/**/*.{js,ts}\"",
+    "lint:fix": "eslint \"src/**/*.{js,jsx,ts,tsx}\" \"scripts/**/*.{js,ts}\" --fix",
     "lint-staged": "lint-staged",
     "commitlint": "commitlint",
-    "coveralls": "npm run test:ci -- --coverage --watchAll=false && cat ./coverage/lcov.info"
+    "coveralls": "npm run test:ci -- --coverage --watchAll=false && cat ./coverage/lcov.info",
+    "translation:check": "ts-node-script scripts/checkMissingTranslations.ts"
```
jackcmeyer, line 117: I think that we should run this script when `yarn start` and `yarn test` are run.

scripts/checkMissingTranslations.ts
```diff
@@ -0,0 +1,59 @@
...
+    return
+  }
+  const defaultKeys: string[] = Object.keys(defaultLanguageObject)
+  const translatedKeys: string[] = Object.keys(translatedLanguageObject)
+  if (defaultKeys.length === 0 || translatedKeys.length === 0) {
+    return
+  }
+  defaultKeys.forEach((key) => {
+    if (!translatedLanguageObject[key]) {
+      console.log(`📙 The key ${key} is not present for path ${searchingPath}`)
```
jackcmeyer, line 24: I think that we should make this a `console.warn()`. Also, it might be helpful to add some color to these log messages using something like https://www.npmjs.com/package/chalk.

scripts/checkMissingTranslations.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const checkRecursiveTranslation = (
+  searchingPath: string,
+  defaultLanguageObject: ResourceKey,
+  translatedLanguageObject: ResourceKey,
+) => {
+  if (typeof defaultLanguageObject === 'string' || typeof translatedLanguageObject === 'string') {
+    if (typeof defaultLanguageObject === 'object') {
+      console.log(
+        `📙 Found a string for path ${searchingPath} while is and object for the default language`,
```
jackcmeyer, line 12: I don't understand what this log message is trying to convey. Could you please explain?

scripts/checkMissingTranslations.ts
jackcmeyer, line 24: I'm thinking that we also add the language code to the log message.. something like 

`[${languageCode}]: No translation present for ${key} in ${searchPath}`

marcosvega91: Yes I though too, but I didn’t want to add an external library only for this sonI used emoji

marcosvega91: Yes you’re right

jackcmeyer: It will only be a devDepedency, so I don't think adding an external package is a bad thing. We'll just want to make sure that it doesn't have a restrictive license and is relatively stable.

package.json
marcosvega91: Yes we can

scripts/checkMissingTranslations.ts
marcosvega91: Yes of course. I’ll fix this

scripts/checkMissingTranslations.ts
marcosvega91: This statement is the default case to make the recursive works and It will cover the following case
default translation
```json
{
  "translations":{
    "patient":{
      "errors":{
        "foo":"bar"
      }
    }
  }
}
```
comparing translation

```json
{
  "translations":{
    "patient":{
      "errors":"foo"
    }
  }
}
```
Now image to call checkRecursiveTranlation with
1) searchPath = translations->patient->errors
2) `{"foo":"bar"}`
3) `"foo"`

The comparing translation has an issue because it is directly a string instead of an object

package.json
fox1t: I don't like adding ts-node-script as new dependency. Can we manage this in some other way?

fox1t:
Thanks for you contribution @marcosvega91!

marcosvega91:
> let's remove ts-node-scripts in favor of a precompiled bin file.

Hey @fox1t you mean to generate js file from tsc command?

Regards 
Marco

package.json
tehKapa: @marcosvega91 any news about this?

marcosvega91: I asked more info about this in the other comment of @fox1t. I don't understand if he wants to keep .ts script and add a new command to convert to .js and run it or do something else

fox1t: @marcosvega91 I am really sorry I completely missed you response! I think we can add bin folder that contains compiled script and use node to run it. Basically add this file to the standard compilation done by our tsconfig.json. What do you think about it @marcosvega91 ?

marcosvega91: I have updated the script. Now ts file is compiled to js and then run as node script

fox1t:
Awesome!
LGTM

marcosvega91:
I see that something was wrong with build. I fixed that

PR #2035, opened 2020-05-02
feat(i18n): add italian translation

marcosvega91:
I added italian translation because was not present.
Some member of the core team are italian, so why not add it? :)

- italian translation

_Note: pull requests without proper descriptions may simply be closed without further discussion. We appreciate your contributions, but need to know what you are offering in clearly described format. Provide tests for all code that you add/modify. If you add/modify any components update the storybook. Thanks! (you can delete this text)_


tehKapa:
Hi @marcosvega91, thanks for your translation! 

Italian localization has never been added because we wanted to focus on the most common languages used in the places where HR is actually used (Africa, Asia and India). Your translation is still very much appreciated :D

src/locales/it/translations/actions/index.ts
```diff
@@ -0,0 +1,15 @@
+export default {
+  actions: {
+    label: 'Azioni',
+    edit: 'Modifica',
+    save: 'Salva',
+    update: 'Aggiorna',
+    complete: 'Completo',
+    delete: 'Rimuovi',
+    cancel: 'Cancella',
```
tehKapa, line 9: It would be more accurate to translate it into "Annulla"

src/locales/it/translations/actions/index.ts
```diff
@@ -0,0 +1,15 @@
...
+    new: 'Nuovo',
+    list: 'Lista',
+    search: 'Cerca',
+    confirmDelete: 'Cancella conferma',
```
tehKapa, line 13: It would be more accurate to translate it into "Conferma cancellazione" or "Conferma eliminazione"

src/locales/it/translations/labs/index.ts
```diff
@@ -0,0 +1,29 @@
+export default {
+  labs: {
+    label: 'Laboratori',
+    requests: {
+      label: 'Richieste di laboratorio',
+      new: 'Nuova richiesta di laboratorio',
+      view: 'Visualizza richiesta',
+      cancel: 'Cancella',
```
tehKapa, line 8: It would be more accurate to translate it into "Annulla"

src/locales/it/translations/actions/index.ts
marcosvega91: You are totally right, why I wrote 'Cancella conferma'.... doesn't make sense

src/locales/it/translations/labs/index.ts
marcosvega91: mmm are you sure? In this case 'Cancella' stands for 'Change the state to Cancellato'.
In english is the same as Delete. 

'Annulla' is better when you need to close a modal, or block some operation. What do you think?

irvelervel: I don't think it's the same as Delete in this case. It means to cancel a lab request. Imho it should be "Annulla richiesta".

src/locales/it/translations/actions/index.ts
```diff
@@ -0,0 +1,15 @@
+export default {
+  actions: {
+    label: 'Azioni',
+    edit: 'Modifica',
+    save: 'Salva',
+    update: 'Aggiorna',
+    complete: 'Completo',
```
irvelervel, line 7: This should be "Completa" in my opinion. It's an action, not an adjective.

src/locales/it/translations/labs/index.ts
```diff
@@ -0,0 +1,29 @@
...
+    },
+    lab: {
+      status: 'Stato',
+      for: 'Per',
+      type: 'Tipo',
+      result: 'Risultato',
+      notes: 'Note',
+      requestedOn: 'Richiesta fatta il',
+      completedOn: 'Completato il',
+      canceledOn: 'Cancellato il',
```
irvelervel, line 25: Same reasons as above, `Annullata il`.

src/locales/it/translations/labs/index.ts
marcosvega91: Ok, I changed to "Annulla richiesta"

PR #2074, opened 2020-05-13
refactor(linter): Add linter rules for import order

MatthewDorner:
There are lots of modules with lots of imports, but no linter rules for ordering the imports, so here is my proposal for adding them.

This is a matter of opinion on coding style, so let me know what you think. A more controversial choice I made is to disallow all relative imports. We currently have something (tsdx? I'm not sure) that allows internal modules to be imported with absolute paths, such as `import Dashboard from 'dashboard/Dashboard'`, etc. But right now about half the imports use absolute and half use relative paths. I like the idea of only allowing absolute or relative paths. So here, I enforced only absolute paths.

With these settings, ordering of imports is auto-fixable but if you try to enter a relative path import you have to fix it manually.

- Add `import/order` rule to enforce a newline between external and internal imports, and to alphabetize imports in each of those two groups.
- Add `no-restricted-imports` rule to disallow relative imports.
- Stop exporting the repository classes from `AppointmentRepository`, `PatientRepository`, etc. This is because it violates [no-named-as-default](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-named-as-default.md). Since there is nothing in the code that imports the actual classes (we default export an instance of the class instead), I just removed the exports.
- Also turned off `import/no-cycle`. This and `no-named-as-default` must have started getting checked now that I added the import plugin. It looks like we still have some import cycles centering around the Redux store and reducers. They don't seem to be causing any bugs now but could be something to look into.

- https://bundlephobia.com/result?p=eslint-import-resolver-typescript@2.0.0 (dev dependency)


fox1t:
Hi @MatthewDorner! Thanks (again :) ) for your contribution. However i strongly dislike the use of tsconfig.json that lets relative imports act like absolute ones. They are harder to maintain and are not standard across all of editors and environments. On the other hand, i strongly agree with you about other rules you enforced. (internal/external, no-named-as-default and import/no-cycle).
This changes has to be propagated also to all other repositories.

Can you make me some examples why you think it is harmful having both absolute and relative?

fox1t:
@MatthewDorner this is the [line](https://github.com/HospitalRun/hospitalrun-frontend/blob/master/tsconfig.json#L32) that allows importing relative as absolute and I suggest to just remove it since we already have pretty much of confusion when new contributors try to add code.

MatthewDorner:
@fox1t I agree with removing the line you linked so only relative is allowed. I just didn't know how that option was set.

I think having the option of both relative and absolute is confusing. Also, if a single file has half the internal imports use a relative path and the other half absolute (as many of the files in this repo currently do), it defeats the purpose of alphabetizing them. So I just think it should be one or the other, I don't have a preference for which. But going with relative only will make it easier to apply this rule in `components` as well.

fox1t:
> @fox1t I agree with removing the line you linked so only relative is allowed. I just didn't know how that option was set.
> 
> I think having the option of both relative and absolute is confusing. Also, if a single file has half the internal imports use a relative path and the other half absolute (as many of the files in this repo currently do), it defeats the purpose of alphabetizing them. So I just think it should be one or the other, I don't have a preference for which. But going with relative only will make it easier to apply this rule in `components` as well.

OK nice! Then let's remove that line from tsconfig.json and thanks a lot for pointing this out, i completely missed it! :/

MatthewDorner:
I think it looks a lot better with only the relative paths. I can do `components` library next.

PR #2085, opened 2020-05-17
fix(incidents): add loading during fetch phase

### LINKED ISSUE #2081: Unable to view an incident
## 🐛 Bug Report

I cannot view an incident. 

## To Reproduce
1. Create an incident
2. Navigate to the incidents list
3. Refresh the browser
4. Click on an incident and see the app never load the incident

See error in console: 
```
react-dom.production.min.js:209 RangeError: Invalid time value
    at D (index.js:371)
    at Te (ViewIncident.tsx:41)
    at Gi (react-dom.production.min.js:153)
    at vc (react-dom.production.min.js:261)
    at us (react-dom.production.min.js:246)
    at cs (react-dom.production.min.js:246)
    at Qc (react-dom.production.min.js:239)
    at react-dom.production.min.js:123
    at scheduler.production.min.js:19
    at Uo (react-dom.production.min.js:122)
```

## Expected behavior
I would expect it to load an incident.

marcosvega91:
fix #2081

- Added loading text until the incident if fetched

### DIFF
--- a/src/incidents/view/ViewIncident.tsx
+++ b/src/incidents/view/ViewIncident.tsx
@@ -1,4 +1,4 @@
-import { Column, Row } from '@hospitalrun/components'
+import { Column, Row, Spinner } from '@hospitalrun/components'
 import format from 'date-fns/format'
 import React, { useEffect } from 'react'
 import { useTranslation } from 'react-i18next'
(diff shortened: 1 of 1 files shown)


src/incidents/view/ViewIncident.tsx
```diff
@@ -31,72 +31,74 @@ const ViewIncident = () => {
...
-          <TextFieldWithLabelFormGroup
-            label={t('incidents.reports.description')}
-            name="description"
-            value={incident?.description}
-          />
-        </Column>
-      </Row>
-    </>
-  )
+  if (incident) {
```
jackcmeyer, line 34: We have a loading status as part of the incident slice that we should add into this check as well.

src/incidents/view/ViewIncident.tsx
```diff
@@ -31,72 +31,74 @@ const ViewIncident = () => {
...
+              label={t('incidents.reports.description')}
+              name="description"
+              value={incident?.description}
+            />
+          </Column>
+        </Row>
+      </>
+    )
+  }
+  return <h1>Loading...</h1>
```
jackcmeyer, line 101: We have a loading component that we should use here.

PR #2108, opened 2020-05-29
feat(navbar): add shortcut icon to the create pages

### LINKED ISSUE #2099: Add quick create feature in the navbar
## 🚀 Feature Proposal
In the top right of the navbar, there should should be a feature that allows me to quickly navigate to create pages that I have access to do. 

## Technical Notes
- Icon should be the `add` icon from `@hospitalrun/components`
- Options should be listed in the same order as the sidebar
- Only options that I have access to view should show up

kumikokashii:
Fixes #2099.

- Add a plus icon in the navbar. When clicked, it opens a list of links to pages where the user has a permission to create: patient, appointment, etc.

### DIFF
--- a/src/__tests__/components/Navbar.test.tsx
+++ b/src/__tests__/components/Navbar.test.tsx
@@ -3,23 +3,59 @@ import '../../__mocks__/matchMediaMock'
 import { Navbar as HospitalRunNavbar } from '@hospitalrun/components'
 import { mount } from 'enzyme'
 import { createMemoryHistory } from 'history'
+import _ from 'lodash'
 import React from 'react'
 import { act } from 'react-dom/test-utils'
+import { Provider } from 'react-redux'
 import { Router } from 'react-router-dom'
+import createMockStore from 'redux-mock-store'
+import thunk from 'redux-thunk'
 
 import Navbar from '../../components/Navbar'
+import Permissions from '../../model/Permissions'
+import { RootState } from '../../store'
+
+const mockStore = createMockStore<RootState, any>([thunk])
 
 describe('Navbar', () => {
   const history = createMemoryHistory()
-  const setup = () =>
-    mount(
+
+  const setup = (permissions: Permissions[]) => {
+    const store = mockStore({
+      title: '',
+      user: { permissions },
+    } as any)
+
+    const wrapper = mount(
       <Router history={history}>
-        <Navbar />
+        <Provider store={store}>
+          <Navbar />
+        </Provider>
       </Router>,
     )
+    return wrapper
+  }
 
-  const wrapper = setup()
+  const allPermissions = [
+    Permissions.ReadPatients,
+    Permissions.WritePatients,
+    Permissions.ReadAppointments,
+    Permissions.WriteAppointments,
+    Permissions.DeleteAppointment,
+    Permissions.AddAllergy,
+    Permissions.AddDiagnosis,
+    Permissions.RequestLab,
+    Permissions.CancelLab,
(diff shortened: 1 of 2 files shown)


src/components/Navbar.tsx
```diff
@@ -1,11 +1,53 @@
...
+    if (permissions.includes(page.permission)) {
+      addDropdownList.push({
+        type: 'link',
+        label: page.label,
+        onClick: () => {
+          history.push(page.path)
+        },
+      })
+    }
+  })
```
jackcmeyer, line 49: Feel free to ignore this, but a more functional of writing this could be. I think it may be more readable, but I could be in the minority here. 

```ts
  const addDropdownList: { type: string; label: string; onClick: () => void }[] = addPages
    .filter((page) => permissions.includes(page.permission))
    .map((page) => ({
      type: 'link',
      label: page.label,
      onClick: () => {
        history.push(page.path)
      },
    }))
```

kumikokashii: such a great idea! ✅ in cd9e4bd

src/__tests__/components/Navbar.test.tsx
```diff
@@ -3,23 +3,59 @@ import '../../__mocks__/matchMediaMock'
 import { Navbar as HospitalRunNavbar } from '@hospitalrun/components'
 import { mount } from 'enzyme'
 import { createMemoryHistory } from 'history'
+import _ from 'lodash'
```
jackcmeyer, line 6: prefer the import style of import `{ cloneDeep } from 'loadash/cloneDeep'` that way the whole library doesn't get imported here.

src/__tests__/components/Navbar.test.tsx
```diff
@@ -3,23 +3,59 @@ import '../../__mocks__/matchMediaMock'
...
+    Permissions.RequestLab,
+    Permissions.CancelLab,
+    Permissions.CompleteLab,
+    Permissions.ViewLab,
+    Permissions.ViewLabs,
+    Permissions.ViewIncidents,
+    Permissions.ViewIncident,
+    Permissions.ReportIncident,
+  ]
+  const wrapper = setup(allPermissions)
```
jackcmeyer, line 56: I think it makes more sense to do `const wrapper = setup(allPermissions)` in each `it` test. That way each test doesn't "share" a wrapper.

kumikokashii:
@jackcmeyer yes, the test file was updated with Provider. and 2 new tests were added for the Navbar.

src/__tests__/components/Navbar.test.tsx
```diff
@@ -142,4 +178,24 @@ describe('Navbar', () => {
       expect(children.props.children[1].props.children).toEqual('actions.search')
     })
   })
+
+  describe('add new', () => {
+    let addNew = hospitalRunNavbar.find('.add-new')
+    let children = addNew.first().props().children as any
```
jackcmeyer, line 184: let's make these variables scoped to each `it` test.

src/__tests__/components/Navbar.test.tsx
kumikokashii: ✅ in 860fb8c

src/__tests__/components/Navbar.test.tsx
kumikokashii: ✅ in 860fb8c 🎈

src/__tests__/components/Navbar.test.tsx
kumikokashii: ✅ in 860fb8c

PR #2109, opened 2020-05-30
feat(network-status): Notify users when they're working offline

### LINKED ISSUE #1773: Notify users when they're working offline
The software has no way, as far as I have seen, for users to tell whether they're working offline or online. This could cause users to not notice if a server goes down or if a network connection is severed, potentially for a long time. Also it could lead to data loss if browser data is cleared either by mistake or if a user thinks it's synced with the server but is not, or obviously if a laptop is lost or HDD failure.

I suggest the frontend should display a bar at the top if it's working offline that says something like "OFFLINE: Unable to connect to server to save data. Erasing browser history will result in the loss of unsaved changes." or some other device to convey the same information.

alessioprestileo:
Fixes #1773.

- Implement NetworkStatus component
- Write tests for NetworkStatusMessage component
- Render NetworkStatus component in src/HospitalRun.tsx


src/components/network-status/NetworkStatus.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { useState, useEffect } from 'react'
+
+import { NetworkStatusMessage } from './NetworkStatusMessage'
+
+export const NetworkStatus = () => {
```
kumikokashii, line 5: I wonder if we can tweak this function a bit and turn it into a hook that returns `networkStatus` for its increased flexibility in future use. This would be called `useNetworkStatus`, and `NetworkStatusMessage` would use it by:

`const networkStatus = useNetworkStatus()`

and in `HospitalRun.tsx`, `NetworkStatusMessage` would be used?

alessioprestileo:
@tehKapa 
I don't know what the problem you faced could be caused by, I tried this link
https://hospitalrun-frontend-git-fork-alessioprestileo-master.hospitalrun.now.sh/labs
on Firefox and Chromium and it seems to work correctly.

src/components/network-status/NetworkStatus.tsx
alessioprestileo: Good idea, I'll do that :+1:

alessioprestileo: Actually, I started to implement your suggestion and now I see that, if I use the useNetworkStatus hook inside the NetworkStatusMessage component, it will make it harder to test NetworkStatusMessage, because it won't be depending on any props anymore.
What do you think about using the useNetworkStatus hook in the HospitalRun component and sending the network status object into NetworkStatusMessage?
That way NetworkStatusMessage will still be easy to test.

kumikokashii: @alessioprestileo Ideally, we'll mock the value like this? https://www.richardkotze.com/coding/mocking-react-hooks-unit-testing-jest

alessioprestileo: @kumikokashii Ok, I'll try that :+1:

src/components/network-status/styles.css
```diff
@@ -0,0 +1,18 @@
...
+  z-index: 9999;
+}
+
+.network-status-message.offline {
+  background-color: rgba(255, 0, 0, 0.65);
+}
+
+.network-status-message.online {
+  background-color: rgba(0, 255, 0, 0.55);
+}
```
kumikokashii, line 18: Awesome work on the hook! Thank you 🙏

One more thing I noticed is the use of custom css. To be consistent with the rest of the project, let's use Bootstrap classes and be free from this css file? For example, we want the status message stick to the bottom, which would be accomplished by adding "fixed-bottom" to the div element's className.

src/components/network-status/NetworkStatusMessage.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { useState } from 'react'
+
+import { useNetworkStatus } from './useNetworkStatus'
+import './styles.css'
+
+export const OFFLINE_MESSAGE = 'you are working in offline mode'
+export const ONLINE_MESSAGE = 'you are back online'
```
jackcmeyer, line 7: these messages will need to be internationalized. no need to get all of the translations, but let's reference this as a key and then put these messages in the `locales` folder

tehKapa:
> @tehKapa
> I don't know what the problem you faced could be caused by, I tried this link
> https://hospitalrun-frontend-git-fork-alessioprestileo-master.hospitalrun.now.sh/labs
> on Firefox and Chromium and it seems to work correctly.

Ok, got it. On Safari doesn't work.

kumikokashii:
On page reload (like a refresh), the offline message is absent. I think we'd like to notify the user about the status if they are working offline at the page load. This will be nice when user is offline, and then refreshes a page. The status bar would 'stay' there to indicate user is still offline.

src/components/network-status/styles.css
alessioprestileo: @kumikokashii Ok, I'll use Bootstrap

src/components/network-status/NetworkStatusMessage.tsx
alessioprestileo: Ok, I'll do that

alessioprestileo:
> On page reload (like a refresh), the offline message is absent. I think we'd like to notify the user about the status if they are working offline at the page load. This will be nice when user is offline, and then refreshes a page. The status bar would 'stay' there to indicate user is still offline.

@kumikokashii I don't understand this comment... If the user is offline, then the page cannot be refreshed, the browser will show an error when trying to refresh the page.
Did you mean something different?

tehKapa:
> > On page reload (like a refresh), the offline message is absent. I think we'd like to notify the user about the status if they are working offline at the page load. This will be nice when user is offline, and then refreshes a page. The status bar would 'stay' there to indicate user is still offline.
> 
> @kumikokashii I don't understand this comment... If the user is offline, then the page cannot be refreshed, the browser will show an error when trying to refresh the page.
> Did you mean something different?

No, when a user reloads a page, HR must continue to work (since it is offline first).

Here's a gif of the problem. https://jmp.sh/EvkscMx 
I translate Firefox label: "_Lavora non in linea_" = "_Work offline_"

alessioprestileo:
@tehKapa Ok, now I see your point... I didn't know that there was an option in the browser to work offline, I was testing this feature by turning the wi-fi off.
At the moment the NetworkStatusMessage component is only listening to the event dispatched when the network status changes. When it's mounted it assumes the network is online. I can improve it so that when it's mounted it checks the network status using "navigator.onLine" (https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine).
Is this what you meant @kumikokashii?

kumikokashii:
> When it's mounted it assumes the network is online. I can improve it so that when it's mounted it checks the network status using "navigator.onLine" (https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine).

@alessioprestileo That looks like what we are looking for. Thank you 🙂

alessioprestileo:
My commits from today should address all the comments :smiley: 
Sorry for the force-push, not sure why I needed that :facepalm:

PR #2112, opened 2020-06-02
fix(sidebar): sidebar should only show links user has access to see

### LINKED ISSUE #2110: Sidebar should only show links which user has access to see
## 🐛 Bug Report

Right now, the sidebar shows all links, no matter what the user has access to see based on their permissions. 

Sidebar links should be limited to what the user is allowed to see.

blestab:
Fixes #2110.

- Only add patients/new link if user has Permissions.WritePatients 
- Only add patients list link if user has Permissions.ReadPatients 
- Only add appointments/new link if user has Permissions.WriteAppointments
- Only add appointments list link if user has Permissions.ReadAppointments
- Only add labs/new link if user has Permissions.RequestLab
- Only add labs list link if user has Permissions.ViewLabs
- Only add incidents/new link if user has Permissions.ReportIncident
- Only add incidents list link if user has Permissions.ViewIncidents
- Unit Tests (do not show link if user does not have the appropriate privileges to access the link)


tehKapa:
Let's have @fox1t check it too, since it's something related to ACL server side feature

blestab:
> 
> 
> we should have tests to prove when certain permissions are not on the user, then the links do not render.

Thanks @jackcmeyer, I will attempt to add those in.

blestab:
> 
> 
> > we should have tests to prove when certain permissions are not on the user, then the links do not render.
> 
> Thanks @jackcmeyer, I will attempt to add those in.

I have pushed some unit tests code. 

Note though, I thought I only needed to test those scenarios where we do not want to show links since the existing tests, by virtue of them having all privileges setup, are also therefore testing cases where we do want the links to show, i.e. when the user has the correct privileges to access the link(s) in question; I hope that was an okay assumption to make.

src/__tests__/components/Sidebar.test.tsx
```diff
@@ -80,6 +116,14 @@ describe('Sidebar', () => {
       expect(listItems.at(3).text().trim()).toEqual('patients.newPatient')
     })
 
+    it('should not render the new_patient link when the user does not have write patient privileges', () => {
+      const wrapper = setupNoPermissions('/patients')
+
+      const listItems = wrapper.find(ListItem)
+
+      expect(listItems.at(3).text().trim()).not.toEqual('patients.newPatient')
```
kumikokashii, line 124: This is checking only one list item to not equal to 'patients.newPatient'. We want to check none of the list items equal to 'patients.newPatient'. I imagine something like:

```
listItems.forEach((_, i) => {
  expect(listItems.at(i).text().trim()).not.toEqual('patients.newPatient')
})
```

blestab: Okay makes sense, i was wondering how to  make the test more robust. I'll try this approach. Thanks @kumikokashii

PR #2103, opened 2020-05-28
[WIP] feat(patient) secondary, tertiary addresses, phone numbers, and emails

sourabbanka22:
**A Feature that adds the ability to define secondary, tertiary, etc. phone numbers, addresses, and
emails for a patient.**

New Feature #1920 
Fixes #1920 .

Below are some Screenshots of the Added Functionality:
**Before Adding PhoneNumber and Email of a Patient:**
Before

**After Adding PhoneNumber and Email:**
After

User needs to click on the **Add Button** to add phone numbers, emails, etc to the respective array.

Currently I have commented out few tests related to Phone Number, Email and Address Models

**Please let me know if there are any Modifications or Changes required.**

PR #2113, opened 2020-06-03
feat(patient): multiple contact info

### LINKED ISSUE #1920: Secondary, tertiary (etc) addresses, phone numbers, and emails for a patient
## 🚀 Feature Proposal
I would like the ability to define secondary, tertiary, etc. phone numbers, addresses, and emails for a patient. 

When creating a new patient or editing patient, I should be able to add these extra fields. 

It should also be able to label what kind of phone number, address, or email it is. 

Phone numbers can be one of the following: 
- home 
- work 
- temp 
- old
- mobile

Email can be one of the following: 
- home 
- work 
- temp 
- old
- mobile

Addresses can one of the following: 
- home
- work 
- temp 
- old 
- billing

## Motivation
We should be flexible in the type of data that can handle. 

## Example
* Apple Contacts

## Technical notes
* Will require changes to the Patient model to support a list of addresses, phone numbers, emails. 
* Will require changes to some models to support the label of the phone number, address, or email (in some cases it might even require creating new models).

kumikokashii:
Fixes #1920. 
This PR evolved from https://github.com/HospitalRun/hospitalrun-frontend/pull/2103.

**Adding**
When adding or editing a patient, more than one entry of contact info can be added per patient. These are for phone numbers, emails, and addresses.


kumikokashii:
WIP:
- Fix: Removing a blank entry in the middle behaves in a strange way. 
- Add tests

src/patients/ContactInfo.tsx
```diff
@@ -0,0 +1,211 @@
...
+      </Column>
+      <Column className="d-none d-sm-block" sm={8}>
+        {t(label)}
+      </Column>
+    </Row>
+  )
+
+  // todo: acts strange when deleting empty rows above non-empty rows.
+  // suspect TextInputWithLabelFormGroup missing value
+  const entries = data.map((entry, i) => {
```
kumikokashii, line 96: note: might be a good idea to wait till the components library get updated, to see if that'll fix it

src/patients/ContactInfo.tsx
```diff
@@ -0,0 +1,211 @@
...
+    const error = getError(i)
+    return (
+      // todo: want a better key, and possibly name
+      // eslint-disable-next-line react/no-array-index-key
+      <Row key={i}>
```
kumikokashii, line 101: temporary solution to use key = i

scripts/checkMissingTranslations.ts
```diff
@@ -1,4 +1,4 @@
-import chalk from 'chalk'
+import * as chalk from 'chalk'
```
fox1t, line 1: Usually we don't need to use `* as`  syntax since we have `allowSyntheticDefaultImports: true`.
Is there any particular reason you changed this line?

kumikokashii: `npm run start` has  an error like below.

Screen Shot 2020-06-15 at 12 01 08 PM

jackcmeyer: allow synthetics is not allowed in this `/scripts` ts config https://github.com/HospitalRun/hospitalrun-frontend/blob/master/scripts/tsconfig.json

src/patients/ContactInfo.tsx
```diff
@@ -0,0 +1,211 @@
...
+    return ''
+  }
+
+  const typeLabel = t('patient.contactInfoType.label')
+  const typeOptions = Object.values(ContactInfoTypes).map((value) => ({
+    label: t(`patient.contactInfoType.options.${value}`),
+    value: `${value}`,
+  }))
+
+  const addLabel = t('actions.add')
```
jackcmeyer, line 68: i think that this variable can be inlined

src/patients/ContactInfo.tsx
```diff
@@ -0,0 +1,211 @@
...
+                        : { value, type },
+                    )
+                    onChange(newData)
+                  }
+                : undefined
+            }
+          />
+        </Column>
+        <Column sm={8}>
+          {['tel', 'email'].indexOf(type) > -1 ? (
```
jackcmeyer, line 124: i think we can just pass in the type of text input box we wish to display as a prop to contact info component. 

to me it seems like it is less likely to break if we explicitly define it as a prop. Otherwise, the logic for determining whether or not the text box is a phone number or email is based on the types of the labels for the contact information types we support. 

it also has the added benefit of removing the need for the conditional.

src/patients/ContactInfo.tsx
```diff
@@ -0,0 +1,211 @@
...
+    onChange(newData)
+    setTempErrors(newTempErrors)
+  }
+
+  const addButton = (
+    <div className="text-right">
+      <button
+        type="button"
+        className="btn btn-link"
+        onClick={isValid && onChange ? onClickAdd : undefined}
```
jackcmeyer, line 187: the `onClickAdd` function is already checking for `!!isValid || !onChange`, so I believe this ternary is a duplicate.

src/patients/ContactInfo.tsx
```diff
@@ -0,0 +1,211 @@
...
+              feedback={error}
+              isInvalid={!!error}
+            />
+          )}
+        </Column>
+      </Row>
+    )
+  })
+
+  const onClickAdd = () => {
```
jackcmeyer, line 161: I prefer the name `onAddClick`.

src/patients/ContactInfo.tsx
```diff
@@ -0,0 +1,211 @@
...
+  return isEditable && data.length === 0 ? (
+    <Spinner color="blue" loading size={20} type="SyncLoader" />
+  ) : (
+    <div>
+      {data.length > 0 ? header : null}
+      {entries}
+      {isEditable ? addButton : null}
+    </div>
+  )
+}
```
jackcmeyer, line 203: using javascript conditionals, rather than JSX may clean this code up a bit: 


```ts
if(isEditable && data.length === 0) {
	return <Spinner color="blue" loading size={20} type="SyncLoader" />
}

return (
    <div>
      {data.length > 0 ? header : null}
      {entries}
      {isEditable ? addButton : null}
    </div>
  )
```

src/patients/GeneralInformation.tsx
```diff
@@ -148,7 +140,14 @@ const GeneralInformation = (props: Props) => {
...
                 value={`${differenceInYears(new Date(Date.now()), new Date(patient.dateOfBirth))}`}
                 isEditable={isEditable}
-                onChange={onApproximateAgeChange}
+                onChange={
+                  (event) =>
+                    onFieldChange(
+                      'dateOfBirth',
+                      guessDOBfromApproximateAge(event.currentTarget.value),
+                    )
+                  // eslint-disable-next-line react/jsx-curly-newline
```
jackcmeyer, line 149: let's remove the eslint disable, and format to however eslint perferes.

src/patients/ContactInfo.tsx
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    // 1. pick up only non-empty string
+    const newData = data.filter(({ value }) => value.trim() !== '')
+
+    // 2. gather errors
+    const newTempErrors = newData.map(({ value }) => (isValid(value) ? '' : errorMessageLabel))
+
+    // 3. if no error, add a new entry
+    if (newTempErrors.filter((error) => error !== '').length === 0) {
```
jackcmeyer, line 173: this can be simplified by doing `newTempErrors.some((e) => e !== ''))`

src/patients/ContactInfo.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const onClickAdd = () => {
+    if (!isValid || !onChange) {
+      return
+    }
+
+    // 1. pick up only non-empty string
+    const newData = data.filter(({ value }) => value.trim() !== '')
+
+    // 2. gather errors
+    const newTempErrors = newData.map(({ value }) => (isValid(value) ? '' : errorMessageLabel))
```
jackcmeyer, line 170: i'm not sure that this component should do any validation. we should leave that all up to redux.

src/patients/ContactInfo.tsx
```diff
@@ -0,0 +1,211 @@
...
+    if (!onChange) {
+      return
+    }
+
+    // eslint-disable-next-line no-shadow
+    const newData = data.map(({ value, type }) =>
+      value === prevValue ? { value: event.currentTarget.value, type } : { value, type },
+    )
+    onChange(newData)
+  }
```
jackcmeyer, line 80: Let's write this in a way that does shadow variables declared in a higher scope. 

maybe in this case, we shouldn't destructure the param to the function for the map.

src/patients/ContactInfo.tsx
kumikokashii: if we choose to allow a new row to be added every time user clicks on the add button, we'd allow infinitely many rows. that wouldn't be user-friendly.

jackcmeyer: ok, I can see that. Maybe just prevent a new one from being added if there is a blank one then. 

I just hate duplicating validation logic in several spots.

scripts/checkMissingTranslations.ts
fox1t: This is a bug on tsconfig.json side. I am going to fix this.

fox1t: I fixed the tsconfig.json on the master. Can you please merge master in kumikokashii:multiple-contact-info and try again leaving `import chalk from 'chalk'`

fox1t:
I've introduced new conflict on translations. Can you address it? :(

scripts/checkMissingTranslations.ts
kumikokashii: ✅

src/patients/ContactInfo.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const { t } = useTranslation()
+
+  useEffect(() => {
+    if (!onChange) {
+      return
+    }
+    if (data.length === 0) {
+      onChange([...data, { value: '' }])
+    }
+  }, [data, onChange])
```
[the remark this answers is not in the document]
kumikokashii: ✅

src/patients/GeneralInformation.tsx
```diff
@@ -175,7 +174,10 @@ const GeneralInformation = (props: Props) => {
                 label={t('patient.unknownDateOfBirth')}
                 name="unknown"
                 disabled={!isEditable}
-                onChange={(event) => onCheckboxChange(event, 'isApproximateDateOfBirth')}
+                onChange={
+                  (event) => onFieldChange('isApproximateDateOfBirth', event.currentTarget.value)
+                  // eslint-disable-next-line react/jsx-curly-newline
```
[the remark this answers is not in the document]
kumikokashii: It seems like ESLint doesn't have a solution for a case like this? We are unable to proceed without `eslint-disable-next-line`.

src/patients/GeneralInformation.tsx
```diff
@@ -2,48 +2,52 @@ import { Panel, Checkbox, Alert } from '@hospitalrun/components'
...
+      }
+      onChange(newPatient)
     }
+  }
 
-    const approximateDateOfBirth = subYears(new Date(Date.now()), approximateAgeNumber)
-    if (onFieldChange) {
-      onFieldChange('dateOfBirth', startOfDay(approximateDateOfBirth).toISOString())
-    }
+  const guessDOBfromApproximateAge = (value: string) => {
```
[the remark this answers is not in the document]
kumikokashii: ✅

src/patients/patient-slice.ts
```diff
@@ -204,15 +204,33 @@ function validatePatient(patient: Patient) {
...
-  if (patient.email) {
-    if (!validator.isEmail(patient.email)) {
-      error.email = 'patient.errors.invalidEmail'
+  if (patient.emails) {
+    const errors: string[] = []
+    patient.emails.forEach((email) => {
+      if (!validator.isEmail(email.value)) {
+        errors.push('patient.errors.invalidEmail')
+      } else {
+        errors.push('')
```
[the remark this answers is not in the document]
kumikokashii: ✅

src/patients/patient-slice.ts
```diff
@@ -204,15 +204,33 @@ function validatePatient(patient: Patient) {
...
+      }
+    })
+    // Only add to error obj if there's an error
+    if (errors.filter((value) => value === '').length !== patient.emails.length) {
```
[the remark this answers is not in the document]
kumikokashii: ✅

src/patients/ContactInfo.tsx
kumikokashii: ✅

src/patients/ContactInfo.tsx
kumikokashii: ✅

src/patients/ContactInfo.tsx
kumikokashii: ✅

src/patients/ContactInfo.tsx
kumikokashii: ✅

src/patients/ContactInfo.tsx
kumikokashii: ✅

src/patients/GeneralInformation.tsx
kumikokashii: It seems like ESLint doesn't have a solution for a case like this? We are unable to proceed without `eslint-disable-next-line`.

src/patients/ContactInfo.tsx
kumikokashii: Line no longer exists.

src/patients/ContactInfo.tsx
kumikokashii: ✅

src/patients/ContactInfo.tsx
kumikokashii: ✅

kumikokashii:
> I've introduced new conflict on translations. Can you address it? :(

The translation file is working after merging. Thank you 🙏

kumikokashii:
Bug https://github.com/HospitalRun/hospitalrun-frontend/issues/2134 filed for the incorrect behavior. Revisit after the next Components update.

src/patients/GeneralInformation.tsx
```diff
@@ -160,12 +156,12 @@ const GeneralInformation = (props: Props) => {
...
-                isInvalid={error?.dateOfBirth}
                 maxDate={new Date(Date.now().valueOf())}
-                feedback={t(error?.dateOfBirth)}
-                onChange={(date: Date) => {
-                  onDateOfBirthChange(date)
-                }}
+                onChange={(date: Date) => onFieldChange('dateOfBirth', date.toISOString())}
+                isInvalid={!!error?.dateOfBirth}
+                feedback={
+                  error ? (error.dateOfBirth ? t(error.dateOfBirth) : undefined) : undefined
```
jackcmeyer, line 168: we do this sort of login on a lot of spots... lets make a helper hook. 


```
// useTranslation.ts
import { useTranslation } from 'react-i18next'

export default function useTranslator() {
  const { t } = useTranslation()

  function translate(key: undefined): undefined
  function translate(key: string): string
  function translate(key: any): any {
    return key !== undefined ? t(key) : undefined
  }

  return {
    t: translate,
  }
}
```

it would be used like: 


```
const { t } = useTranslator()
```

and we can replace 

`feedback={error ? (error.dateOfBirth ? t(error.dateOfBirth) : undefined) : undefined`}
with
`feedback={t(error?.dateOfBirth)}`

src/patients/edit/EditPatient.tsx
```diff
@@ -69,24 +70,12 @@ const EditPatient = () => {
...
-        {
-          ...patient,
-          fullName: getPatientName(patient.givenName, patient.familyName, patient.suffix),
-          index:
-            getPatientName(patient.givenName, patient.familyName, patient.suffix) + patient.code,
-        },
-        onSuccessfulSave,
-      ),
-    )
+    const newPatient = cleanupPatient(patient)
```
jackcmeyer, line 73: let's have the redux `updatePatient` and `createPatient` do this

src/patients/GeneralInformation.tsx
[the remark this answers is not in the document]
kumikokashii: ✅ This is due to a conflict between eslint and prettier. Proceeding by extracting what's in the bracket into a constant.

src/patients/GeneralInformation.tsx
kumikokashii: ✅ This is due to a conflict between eslint and prettier. Proceeding by extracting what's in the bracket into a constant.

kumikokashii:
> #2134

Fixed before merge.

src/patients/ContactInfo.tsx
```diff
@@ -0,0 +1,149 @@
...
+      // eslint-disable-next-line react/no-array-index-key
+      <Row key={i}>
+        <Column sm={4}>
+          <SelectWithLabelFormGroup
+            name={`${name}Type${i}`}
+            value={entry.type}
+            isEditable={isEditable}
+            options={typeOptions}
+            onChange={
+              onChange
```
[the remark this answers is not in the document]
kumikokashii: ✅

src/patients/ContactInfo.tsx
```diff
@@ -0,0 +1,149 @@
...
+            }
+          />
+        </Column>
+        <Column sm={8}>
+          <Component
+            name={`${name}${i}`}
+            value={entry.value}
+            isEditable={isEditable}
+            onChange={
+              onChange
```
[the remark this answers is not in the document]
kumikokashii: ✅

src/patients/edit/EditPatient.tsx
kumikokashii: ✅

src/patients/GeneralInformation.tsx
kumikokashii: Opened a ticket: https://github.com/HospitalRun/hospitalrun-frontend/issues/2135

PR #2145, opened 2020-06-19
feat(patient): make care plan note not required

### LINKED ISSUE #2127: Care Plan Note should not be required
## 🐛 Bug Report

When creating a care plan, the note field should not be required.

## To Reproduce
1. navigate to patients -> care plan -> add care plan
2. fill in the required fields
3. note should not be required

## Expected behavior
1. note should not be required

## Technical notes
* remote validation from redux thunk
* remove required asterisk

reidmeyer:
Worked with WUSTL Group (Anya Pawar, Samika Kikkeri, Fadel Alkilani). First issue.

fix #2127

- makes care plan note not required

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "@testing-library/react": "~10.1.0",
     "@testing-library/react-hooks": "~3.3.0",
     "@types/enzyme": "^3.10.5",
-    "@types/jest": "~25.2.0",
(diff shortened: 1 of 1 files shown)


jackcmeyer:
Changes look good! Looks like this caused some other tests to fail, however.

PR #2198, opened 2020-07-04
fix(login): improved login validation

### LINKED ISSUE #2182: Username and password should be requred
# 🐛 Bug Report
Username and password should be marked as required fields on the login form and when I try to submit the login form, it should prevent a login request from being made if username and password are not filled in.

## To Reproduce
1. Navigate to the login screen
2. Submit the login form
3. See that username and password are not required

## Expected behavior
* The username and password fields should be marked as required
* When submitting the form, there should be form validation saying username/password is required
* There should be a banner message saying that it was unable to login

### LINKED ISSUE #2183: No message appears when wrong login credentials are entered
# 🐛 Bug Report

When logging in, if wrong credentials are entered, then there should be a message that says: "Unable to login. Incorrect username or password".

## To Reproduce
1. Enter incorrect username or password
2. Click the login button
3. See that no message appears stating that the wrong credentials were entered

## Expected behavior
* There should be a message that says: "Unable to login. Incorrect username or password".

linus345:
Fixes #2182 
Fixes #2183 .

- Marked username and password fields as required.
- Shows banner with error message when username or password is not provided.
- Shows banner with error message when wrong credentials are provided.


tehKapa:
Thanks for the PR. I tried the enhancements that you reported, but only two work:
- Marked username and password fields as required.
- Shows banner with error message when username or password is not provided.

I can’t view any error with wrong credentials

linus345:
Yes it doesn't seem to work in the live preview but it works locally (after setting everything up according to [CONTRIBUTING.md](https://github.com/HospitalRun/hospitalrun-frontend/blob/master/.github/CONTRIBUTING.md#setting-up-the-local-development-environment)). From what I can see in the `network` tab in chrome devtools the live preview doesn't connect to the CouchDB instance which, if it doesn't, is the reason why the last enhancement doesn't work in the live preview. When testing it locally I see a request to `http://localhost:5984/_session` (CouchDB instance) when signing in with both the username and password fields filled in. If the live preview actually is connected to a CouchDB instance I'll take a deeper look of what the problem could be.

src/user/user-slice.ts
```diff
@@ -89,8 +89,10 @@ export const login = (username: string, password: string): AppThunk => async (di
       }),
     )
   } catch (error) {
-    if (error.status === '401') {
-      dispatch(loginError('user.login.error'))
+    if (!username || !password) {
+      dispatch(loginError('Missing required fields.'))
```
jackcmeyer, line 93: error messages need to be internationalized.

src/login/Login.tsx
```diff
@@ -68,6 +68,9 @@ const Login = () => {
               name="username"
               value={username}
               onChange={onUsernameChange}
+              isRequired
+              isInvalid={!!loginError && !username}
+              feedback={!!loginError && !username ? 'Username is required.' : undefined}
```
jackcmeyer, line 73: I think that we should move the "login error" message, to be an object that looks like: 

{
	message: 'some error message',
	username: 'some login error message specific to username'
	password: 'some login error messages specific to password'
}

This is consistent with the rest of the application, and I think simplifies the ternaries since
`isInvalid = error?.username` and feedback is `t(error?.username)`

linus345:
Thanks for the feedback!

> it appears that when I enter invalid credentials, I never see the invalid credential message. (chrome v83.0.4103.116).

Did you test it in the live preview when you weren't able to see the invalid credentials message? If so then the issue may be what I described above: 
> Yes it doesn't seem to work in the live preview but it works locally (after setting everything up according to [CONTRIBUTING.md](https://github.com/HospitalRun/hospitalrun-frontend/blob/master/.github/CONTRIBUTING.md#setting-up-the-local-development-environment)). From what I can see in the `network` tab in chrome devtools the live preview doesn't connect to the CouchDB instance which, if it doesn't, is the reason why the last enhancement doesn't work in the live preview. When testing it locally I see a request to `http://localhost:5984/_session` (CouchDB instance) when signing in with both the username and password fields filled in. If the live preview actually is connected to a CouchDB instance I'll take a deeper look of what the problem could be.

PR #2227, opened 2020-07-11
chore(init): add curl command to create a new DB user in the docker-compose

### LINKED ISSUE #2148: Improve the docker-compose to start CouchDB
It is necessary to add curl command to create a new DB user directly in the docker-compose.
Currently, frontend dev must launch it in manual mode after database startup with this request:

```
curl -X PUT http://admin:password@localhost:5984/_users/org.couchdb.user:username -H "Accept: application/json" -H "Content-Type: application/json" -d '{"name": "username", "password": "password", "metadata": { "givenName": "John", "familyName": "Doe"}, "roles": [], "type": "user"}'
```

The first implementation tests showed CouchDB error
`{"error": "bad_request", "reason": "Missing request body"}`

Link #2137

LoicB:
In the docker-compose file, add a fourth curl command to create a new DB user

Fixes #2148

- add a curl command to create a default user

N/A

I managed to reproduce the issue of having the error message in the logs:
`{"error": "bad_request", "reason": "Missing request body"}`
Then I fixed it by changing the curl command. 

- Now it creates a user, the same way than in the ticket. Is that correct?

- In the manual, there is another curl command, does it need to be included as well?

- Does it need to handle error cases (if the docker-compose command is run twice, the second time curl commands seems to fail)?


tehKapa:
Hi @LoicB, thanks for your submission!

Yes, you could also add the command to make the database public. For error handling, I don't know if you can check inside "command" in the docker-compose. It would be to verify the existence of user record and then skipping all the curl commands. What do you think?

tehKapa:
@LoicB to handle errors (so use conditions statement) the only thing I found is this https://github.com/docker/compose/issues/4638#issuecomment-287930367
so outside of "just running a docker-compose"

tehKapa:
with my last commit I also added the other curl command.

PR #2235, opened 2020-07-13
feat(patients): add no patients exist

### LINKED ISSUE #2231: Implement "no patients" state on the /patients route
## 🚀 Feature Proposal
Implement "no patients" state on the `/patients` route. When there are no patients in the database, then a page that looks like the following should show: 

image

## Technical Notes
- for now, we can use the already existing patients icon in place of the icon that is shown. 
- note: this is state is different than no patients returned via a search.

Fibii:
Fixes  #2231

- add "no patients exists" state when no patients exist in `/patient` as discussed in #2231 
- add test for said changes
- add few locales called `noPatients` for languages: eng/de/fr/ar/ru


Fibii:
so for some reason i added a patient to the vercel build to check if the "add new patient" shows up when patients exist (even tho the tests passed) and i couldn't find a way to remove the patient, which means my commit won't be displayed (because there's a patient) so whomever reviews the code in vercel have to remove that test patient.

src/patients/list/ViewPatients.tsx
```diff
@@ -91,6 +102,34 @@ const ViewPatients = () => {
     setSearchText(event.target.value)
   }
 
+  if (patients && patients.length === 0) {
+    return (
+      <div style={{ display: 'flex', justifyContent: 'center' }}>
+        <div style={{ display: 'flex', flexDirection: 'column' }}>
+          <div style={{ display: 'flex', justifyContent: 'center' }}>
+            <Icon icon="patient" outline={false} size="6x" />
```
jackcmeyer, line 110: lets use the `patients` icon rather than `patient`.

src/patients/list/ViewPatients.tsx
```diff
@@ -91,6 +102,34 @@ const ViewPatients = () => {
...
+          <div style={{ display: 'flex', justifyContent: 'center' }}>
+            <div style={{ textAlign: 'center', width: '60%', margin: 16 }}>
+              <Typography variant="h5">{t('patients.noPatients')}</Typography>
+            </div>
+          </div>
+          <div style={{ display: 'flex', justifyContent: 'center' }}>
+            <Button
+              key="newPatientButton"
+              outlined
+              color="success"
```
jackcmeyer, line 121: rather than using a success and outlined button, I think a `primary` button makes more sense here since it is the call to action on the page.

src/patients/list/ViewPatients.tsx
```diff
@@ -48,22 +58,23 @@ const ViewPatients = () => {
...
-        outlined
-        color="success"
-        icon="patient-add"
-        onClick={() => history.push('/patients/new')}
-      >
-        {t('patients.newPatient')}
-      </Button>,
-    ])
-
+    if (patients && patients.length > 0) {
```
jackcmeyer, line 61: this `patients` variable represents the current patients in state. 

This could be the default page of patients or a search result. When searching for patients and my search results return 0, I would not expect the new icon to appear. 

I think that we should implement a new function called `count` in `Repository` that returns the number of documents that are available for a type. 

Then here, we use that `count` function to determine if the new icon should display or not.

src/patients/list/ViewPatients.tsx
```diff
@@ -91,6 +102,34 @@ const ViewPatients = () => {
...
+              outlined
+              color="success"
+              icon="patient-add"
+              onClick={() => history.push('/patients/new')}
+            >
+              {t('patients.newPatient')}
+            </Button>
+          </div>
+        </div>
+      </div>
```
jackcmeyer, line 129: I think that we should move this to its own component to: 
- keep components smaller and easier to understand
- make it easier to test

jackcmeyer:
@Fibii please let me know if you have any questions or need help addressing the code review feedback.

src/patients/list/ViewPatients.tsx
Fibii: i get it, what i don't get is why do we need to implement `count`? 

it seems like `PatientRepository.findAll()` returns all the existing patients in the db, we can use it as:
1. implement a state called `count`
2. at initial load, setCount(PatientRepository.findAll().length)
3. if (count == 0) show `addNewPations` component, else show the table

but it seems that implementing a `count` in `Repository` seems like a better solution here, since it makes `Repository` more flexible for future issues/enhancements

Fibii:
@jackcmeyer how would i add `count` to `patient-slice`, should i even do that ?

i already added `count` to `Repository` as 
```js
async count(): Promise<number> {
    const result = await this.findAll()
    return result.length
  }
```
what i want to do now is use it in `ViewPatients.tsx` as 
```js
  const { patients, isLoading, count } = useSelector((state: RootState) => state.patients)
```

but i couldn't figure it out. I updated the interface in `patients-slice.ts` to be
```js
interface PatientsState {
  isLoading: boolean
  patients: Patient[]
  count: number
}
```
and called `Repository#count` in `fetchPatients`
```js
export const fetchPatients = (sortRequest: SortRequest = Unsorted): AppThunk => async (
  dispatch,
) => {
  dispatch(fetchPatientsStart())
  const patients = await PatientRepository.findAll(sortRequest)
  const count = await PatientRepository.count()
  // dispatch(fetchPatientsSuccess(... ? ))
}
```
but i couldn't figure what to dispatch there. I need a little bit of help here.

tehKapa:
@fox1t some advice?

src/__tests__/patients/list/ViewPatients.test.tsx
```diff
@@ -80,6 +82,30 @@ describe('Patients', () => {
...
+      const buttonIcon = button.prop('icon')
+      const buttonText = button.prop('children')
+
+      expect(iconType).toEqual('patients')
+      expect(iconSize).toEqual('6x')
+      expect(typographyText).toEqual('patients.noPatients')
+      expect(typographyVariant).toEqual('h5')
+      expect(buttonIcon).toEqual('patient-add')
+      expect(buttonText).toEqual('patients.newPatient')
+    })
```
jackcmeyer, line 107: This test is testing behavior inside the `AddNewPatient` component. I think instead, this test should test that the `AddNewPatient` component gets rendered and then this current test should be moved to an `AddNewPatient` test.

src/patients/view/AddNewPatient.tsx
```diff
@@ -0,0 +1,38 @@
+import { Button, Icon, Typography } from '@hospitalrun/components'
+import React from 'react'
+import { useHistory } from 'react-router-dom'
+
+import useTranslator from '../../shared/hooks/useTranslator'
+
+const AddNewPatient = () => {
```
jackcmeyer, line 7: Thoughts on `NoPatientsExist` for this component name? `AddNewPatient` could potentially be confusing with `NewPatient` component.

PR #2236, opened 2020-07-14
feat(navbar): display currently logged in user name

### LINKED ISSUE #2225: Display currently logged in user name in profile dropdown
## 🚀 Feature Proposal
The currently logged in user should display in the profile icon dropdown in the top right corner. 

The first item in the dropdown should read
"Currently signed in as **${user.givenName} ${user.familyName}**"

Clicking on this dropdown should bring the user to the settings page.

## Motivation
Easily identify who the currently logged in user is

AlexTan331:
Fixes #2225 .

- add ability to display user name in profile dropdown
- add ability to navigate to the setting route
- add new test for display user name 
- add translation key to display profile message

**No new dependency added**

### DIFF
--- a/src/__tests__/shared/components/navbar/Navbar.test.tsx
+++ b/src/__tests__/shared/components/navbar/Navbar.test.tsx
@@ -170,13 +170,22 @@ describe('Navbar', () => {
   })
 
   describe('account', () => {
-    it('should render an account link list', () => {
+    it("should render a link with the user's identification", () => {
       const wrapper = setup(allPermissions)
       const hospitalRunNavbar = wrapper.find(HospitalRunNavbar)
       const accountLinkList = hospitalRunNavbar.find('.nav-account')
       const { children } = accountLinkList.first().props() as any
 
-      expect(children[0].props.children).toEqual([undefined, 'settings.label'])
+      expect(children[0].props.children).not.toBeUndefined()
+    })
+
+    it('should render a setting link list', () => {
+      const wrapper = setup(allPermissions)
+      const hospitalRunNavbar = wrapper.find(HospitalRunNavbar)
+      const accountLinkList = hospitalRunNavbar.find('.nav-account')
+      const { children } = accountLinkList.first().props() as any
+
+      expect(children[1].props.children).toEqual([undefined, 'settings.label'])
     })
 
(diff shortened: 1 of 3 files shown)


src/shared/locales/enUs/translations/user/index.ts
```diff
@@ -13,6 +13,7 @@ export default {
           required: 'Password is required.',
         },
       },
+      success: 'Currently signed in as ',
```
jackcmeyer, line 16: Let's use `currentlySignedInAs` as the key instead of `success`. I think that it makes it more clear what the key is supposed to represent.

src/__tests__/shared/components/navbar/Navbar.test.tsx
```diff
@@ -170,13 +170,22 @@ describe('Navbar', () => {
   })
 
   describe('account', () => {
-    it('should render an account link list', () => {
+    it("should render a link with the user's identification", () => {
```
jackcmeyer, line 173: There should be a test to make sure that it renders the user's first and last name

src/shared/components/navbar/Navbar.tsx
```diff
@@ -98,6 +98,13 @@ const Navbar = () => {
           type: 'link-list-icon',
           alignRight: true,
           children: [
+            {
+              type: 'link',
+              label: `${t('user.login.success')}${user?.givenName} ${user?.familyName}`,
```
jackcmeyer, line 103: I think we should bold the text for the user's name

src/shared/locales/enUs/translations/user/index.ts
jackcmeyer: Let's leave off the trailing space in the key and make whoever uses this key put a space in.

src/shared/components/navbar/Navbar.tsx
AlexTan331: Since whatever put in the label tag will be converted into string, the bold method won't work in this case. What should I do to make sure that only the username part is rendered as bold in the label tag? 
Could anyone give me a hint?

jackcmeyer: You're right. There is no way to do this right now. This feedback can be ignored and we'll open some more issues to get this functionality added.

PR #2249, opened 2020-07-23
feat(medications): implement basic medication module

### LINKED ISSUE #2229: implement basic medication module
## 🚀 Feature Proposal

This issue is to create a very basic medication module. When this issue is finished, I would expect to be able to: 

- create a new medication request
- view a list of all medication requests
- view a single medication request

## Technical Notes
A medication request is defined as:

```
{
  id: string
  requestedBy: string // id of the user who created the request
  requestedOn: string // date when the medication was requested
  status: active | on hold | cancelled | completed | entered in error | stoped | draft | unknown
  intent: proposal | plan | order | original order | reflex order | filler order | instance order | option
  priority: routine | urgent | asap | stat
  medication: string
  patient: string // the id of the patient the request is for
  note: string
  quantity: { value: number, unit: string }
}
```

- There should be a text box which allows me to search patients
- id, requestedBy, requestedOn should be autogenerated

- The table should show: 
requested date, patient name, medication, quantity, status

- Add new set of permissions for each viewing and making medication requests
- Add sidebar link
- Add "Request Medication" to quick create button in navbar

blestab:
Fixes #2229 .

- Add New Medication request
- View/Update Medication request details
- Complete Medication request
- Cancel Medication request
- List Medication requests
- Medications Unit Tests 

None


src/medications/MedicationOptionFields.tsx
```diff
@@ -0,0 +1,47 @@
+import { Option } from '../shared/components/input/SelectWithLableFormGroup'
+import useTranslator from '../shared/hooks/useTranslator'
+
+const MedicationFieldsOptions = () => {
```
blestab, line 4: Currently not used as I am not sure how best to implement this. I wanted a central place where options used in NewMedication and ViewMedication can be stored but when i had this not as a component then i could not get the translation to work since it needs to happen inside a component (i hope that makes sense). Doing translation outside of a component seems to have potential issues? https://github.com/i18next/react-i18next/issues/909

blestab: I have removed this for now, we can consider refactoring in future.

src/__tests__/medications/Medications.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+          },
+        } as any)
+
+        const wrapper = mount(
+          <Provider store={store}>
+            <MemoryRouter initialEntries={['/medications/new']}>
+              <Medications />
+            </MemoryRouter>
+          </Provider>,
+        )
```
jackcmeyer, line 51: this sort of setup is duplicated a lot in this file. To remove some duplication, but to allow for tests to be independent, I think we should make a `setup` function. 

This setup function may look something like: 

```ts
const setup = () => {
        const store = mockStore({
          title: 'test',
          user: { permissions: [Permissions.RequestMedication] },
          breadcrumbs: { breadcrumbs: [] },
          components: { sidebarCollapsed: false },
          medication: {
            medication: ({ id: 'medicationId', patientId: 'patientId' } as unknown) as Medication,
            patient: { id: 'patientId', fullName: 'some name' },
            error: {},
          },
        } as any)

        const wrapper = mount(
          <Provider store={store}>
            <MemoryRouter initialEntries={['/medications/new']}>
              <Medications />
            </MemoryRouter>
          </Provider>,
        )
	return { wrapper: wrapper as ReactWrapper, store }
}

For setup values that change across file the file, we can pass those as parameters to the function.

The goal is to eliminate any "global" setup such as mocks or initialization of variables. The setup function should be a "pure" function. It should take in an arguments that are needed, setup the variables, and then return any values that will be tested.

File: src/__tests__/medications/ViewMedication.test.tsx [hunk 3]
```diff
@@ -0,0 +1,380 @@
...
+    priority: 'routine',
+    quantity: { value: 1, unit: 'unit' },
+    notes: 'medication notes',
+    requestedOn: '2020-03-30T04:43:20.102Z',
+  } as Medication
+
+  let setButtonToolBarSpy: any
+  let titleSpy: any
+  let medicationRepositorySaveSpy: any
+  const expectedDate = new Date()
```
jackcmeyer, line 43: I think we should move the spies and history into the setup function for initialization. if they are tested later, then we should return them from the setup function.

blestab: Thanks @jackcmeyer i have refactored the duplicated code into a setup function where possible. May you take a look if it is now alright or if i may need to take another look.

PR #1995, opened 2020-04-19
feat(labs): use redux 

jackcmeyer:
Adds redux for requesting, canceling, completing, and updating a lab

PR #2273, opened 2020-08-05
refactor(redux): add react-query for incidents and patients search

jackcmeyer:
Fixes #1995 

# TODO: 
- [ ] patients
- [ ] scheduling
- [x] incidents
- [ ] imaging
- [ ] medication
- [ ] labs


fox1t:
This is huge! <3

src/App.tsx
```diff
@@ -33,14 +34,17 @@ const App: React.FC = () => {
...
+    <>
+      <BrowserRouter>
+        <Suspense fallback={<Spinner color="blue" loading size={[10, 25]} type="ScaleLoader" />}>
+          <Switch>
+            <Route exact path="/login" component={Login} />
+            <Route path="/" component={HospitalRun} />
+          </Switch>
+        </Suspense>
+      </BrowserRouter>
+      <ReactQueryDevtools initialIsOpen={false} />
```
fox1t, line 46: Is it a way of making this available only in "dev mode"? I am not fammiliar with react-query-devtools at all.

fox1t:
All seems more clear to me when comes to "fetching" data.

src/App.tsx
blestab: > 
> 
> Is it a way of making this available only in "dev mode"? I am not fammiliar with react-query-devtools at all.

I thought so too then checked its npm page and it says:

_from the npm page React Query Devtools are not imported and used when process.env.NODE_ENV === 'production', so you don't need to worry about excluding them during a production build._

tehKapa:
LGTM though coveralls alert

PR #2289, opened 2020-08-13
chore: remove redux part II

jackcmeyer:

- refactors care plan to use react-query instead of redux
- fixes bug with allergies list not refreshing after allergy had been added


src/patients/care-plans/AddCarePlanModal.tsx
```diff
@@ -38,14 +39,19 @@ const AddCarePlanModal = (props: Props) => {
...
-    dispatch(addCarePlan(patient.id, carePlan as CarePlan))
-  }
-
   const onClose = () => {
     onCloseButtonClick()
   }
 
+  const onSaveButtonClick = async () => {
+    try {
+      await mutate({ patientId: patient.id, carePlan: carePlan as CarePlan })
```
fox1t, line 48: I was wondering do we really need these explicit castings? Using them prevents us from relying on the compiler inference and could move problems and type errors at the runtime.

jackcmeyer: good point, I've refactored types to be more reflective of the data that is expected. I didn't feel that it was necessary to name them so I just inlined them.

PR #2309, opened 2020-08-15
feat(imaging): link image with visit

### LINKED ISSUE #2291: Link imaging request with visit
## 🚀 Feature Proposal

An imaging request should be able to be linked to a visit. When creating an imaging request, a list of visits for the selected patient should show.

AlexTan331:
Fixes #2291 .

- create a dropdown list for visits 
- render a list of visits for the selected patient when requesting a new imaging


src/__tests__/imagings/requests/NewImagingRequest.test.tsx
```diff
@@ -204,12 +213,18 @@ describe('New Imaging Request', () => {
         onChange({ currentTarget: { value: expectedImaging.type } })
       })
 
-      const statusSelect = wrapper.find(SelectWithLabelFormGroup)
+      const statusSelect = wrapper.find('.imaging-status').find(SelectWithLabelFormGroup)
```
blestab, line 216: Just curious why you decided to use the class here? This test would have to be changed if for some reason the class name changes

src/labs/requests/NewLabRequest.tsx
```diff
@@ -121,7 +121,7 @@ const NewLabRequest = () => {
         <div className="row float-right">
           <div className="btn-group btn-group-lg mt-3">
             <Button className="mr-2" color="success" onClick={onSave}>
-              {t('actions.save')}
+              {t('labs.requests.save')}
```
blestab, line 124: This does not seem to be related to the issue you were working on?

src/shared/locales/enUs/translations/labs/index.ts
```diff
@@ -13,7 +13,8 @@ export default {
     },
     requests: {
       label: 'Lab Requests',
-      new: 'New Lab Request',
+      new: 'Request Lab',
```
blestab, line 16: This does not seem to be related to the issue you were working on?

src/shared/locales/enUs/translations/patient/index.ts
```diff
@@ -146,7 +146,7 @@ export default {
       label: 'Visits',
       startDateTime: 'Start Date',
       endDateTime: 'End Date',
-      type: 'Type',
+      type: 'Visit Type',
```
blestab, line 149: Just curious why we decide to add Visit here? Aren't we already looking at Visit* fields? if we have to do it for one should we maybe not also do it for all of them? i.e. Visit Start Date, Visit End Date etc for consistency? Also not sure if this is related to the issue that this PR is meant to be addressing? I stand corrected should it be.

src/__tests__/imagings/requests/NewImagingRequest.test.tsx
AlexTan331: For my point of view, since there are two fields (Selecting status and visit) using SelectWithLabelFormGroup in the new request imaging page, I need to distinguish the two fields with different div classNames (in this case, imaging-status and imaging-visit) so that it is easier to find the component for testing

src/labs/requests/NewLabRequest.tsx
AlexTan331: Yeah, you are right. This change was related to my previous issue #2301 and it should've been merged into master. I am not sure why the commits are shown up here :(

src/shared/locales/enUs/translations/labs/index.ts
AlexTan331: This change was also related to my previous issue #2301 and it should've been merged into master. I am not sure why the commits are shown up here :(

src/shared/locales/enUs/translations/patient/index.ts
AlexTan331: What I was originally thinking is to create a translation key for the visit type so that it can be used in the new request imaging page 
 `<div className="visits">
         {newImagingRequest.patient && visitOption ? (
                <SelectWithLabelFormGroup
                  name="visit"
                  label={t('patient.visits.type')}` 
image


Maybe it will make more sense to create a new translation key in the imaging locale index.ts file to prevent some confusion

src/imagings/requests/NewImagingRequest.tsx
```diff
@@ -94,19 +114,56 @@ const NewImagingRequest = () => {
...
+              ) : (
+                <SelectWithLabelFormGroup
+                  name="visit"
+                  label={t('imagings.imaging.visit')}
+                  isRequired
+                  isEditable={false}
+                  options={[]}
+                  onChange={(values) => {
+                    onVisitChange(values[0])
+                  }}
```
jackcmeyer, line 160: since `defaultSelected` takes an undefined, I think we can simplify this logic to the following (or something similar to this): 

```
<SelectWithLabelFormGroup
  name="visit"
  label={t('imagings.imaging.visit')}
  isRequired
  isEditable={patient !== undefined}
  options={patient.visits?.map((v) => ({ label: v.type, value: v.id })) || []}
  onChange={(values) => {
 onVisitChange(values[0])
}}
          />
```

src/imagings/requests/NewImagingRequest.tsx
```diff
@@ -46,10 +48,21 @@ const NewImagingRequest = () => {
...
-      ...previousNewImagingRequest,
-      patient: patient.fullName as string,
-    }))
+    if (patient) {
+      setNewImagingRequest((previousNewImagingRequest) => ({
+        ...previousNewImagingRequest,
+        patient: patient.fullName as string,
+      }))
+
+      const visits = patient.visits?.map((v) => ({ label: v.type, value: v.id }))
```
jackcmeyer, line 57: I think the label should be something like `${visitType} at `${startDateTime}` or `${visitType} (${startDateTime)}`

src/imagings/requests/NewImagingRequest.tsx
```diff
@@ -46,10 +48,21 @@ const NewImagingRequest = () => {
...
 
   const onPatientChange = (patient: Patient) => {
-    setNewImagingRequest((previousNewImagingRequest) => ({
-      ...previousNewImagingRequest,
-      patient: patient.fullName as string,
-    }))
+    if (patient) {
+      setNewImagingRequest((previousNewImagingRequest) => ({
+        ...previousNewImagingRequest,
+        patient: patient.fullName as string,
```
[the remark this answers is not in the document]
AlexTan331: If this store as patient id, the `patient` field on the table in requested imagings page would display its patient id instead of full name
image

src/imagings/requests/NewImagingRequest.tsx
```diff
@@ -46,10 +48,21 @@ const NewImagingRequest = () => {
...
+      }))
+
+      const visits = patient.visits?.map((v) => ({ label: v.type, value: v.id }))
+      setVisitOption(visits)
```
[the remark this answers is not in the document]
AlexTan331: I believe we need the visits to be a state variable. Because we can only fetch the visits data inside the `onPatientChange` method using the patient parameter (as we fetch the patient using this line `onSearch={async (query: string) => PatientRepository.search(query)}`). Using state variable ensures that the visits data will be re-rendered whenever the patient field value is changed.

src/shared/locales/enUs/translations/patient/index.ts
blestab: You can keep the translation key in patients since visits are part of the patients module. Maybe just call it _Visit_ rather than _Visit Type_ because visit type makes me think (outpatient/inpatient/emergency etc) where here we are giving the user a list of all visits for this patient regardless of the type of visit it is, what do you think?

AlexTan331: On it

PR #2319, opened 2020-08-20
fix(eslint rule and subsequent changes to the codebase): eslint rule no-console added

### LINKED ISSUE #2307: Disallow console log in ESLint
## 🐛 Bug Report

As a followup to https://github.com/HospitalRun/hospitalrun-frontend/issues/2286 I thought it might be desirable to not allow any log messages in the application using an ESlint rule https://eslint.org/docs/rules/no-console

## Proposed changes 
* Remove any `console` call from the application
* Add a rule in the ESLint config that would make the linter fail if there is a `console` call in the application

Is there any good reason why `console` logs should be kept in the application?

tobireuen:
Fixes #2307 

- add eslint rule 'no-console' to throw errors whenever calls to the console object are made
- remove all calls to the console object OR allow them on file level for specific cases

Newly added dependencies with Bundlephobia links:

NONE


couchdb-cleanup.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/zsh
```
fox1t, line 1:
```suggestion
#!/bin/sh
```

fox1t: Not everyone has zsh installed

couchdb-init.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/zsh
```
fox1t, line 1:
```suggestion
#!/bin/sh
```

fox1t (changes requested):
Thanks for your contribution!

couchdb-cleanup.sh
tobireuen: Oops, I didn't even mean to commit this and couchdb-init.sh. These were just helper scripts for my environment.

Is there a way to remove them from the pull request?

fox1t: Actually I like them :)

Maybe you can fix that and move them in a dedicated folder?

tobireuen: They are quite handy but shouldn't be part of this pull request. I'll go ahead and create another one for that ;)

fox1t: Please do it! :)

fox1t (approved):
LGTM

PR #2320, opened 2020-08-20
feat(new scripts and updated contributing.md): added scripts to init and cleanup couchdb

tobireuen:
Added couchdb-init.sh to initialize couchdb and couchdb-cleanup.sh to cleanup couchdb according to instructions in CONTRIBUTING.md. 
I also updated the instructions in CONTRIBUTING.md to use those scripts.

- encapsulated the calls to docker-compose for initializing and cleaning the database in their own scripts
- changed CONTRIBUTING.md accordingly

NONE

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -18,14 +18,14 @@ root of this project to launch CouchDB. Below are the steps:
 
 1. Start [Docker](https://docs.docker.com/get-docker/). Install it if you don't have it yet.
 2. Install [Docker Compose](https://docs.docker.com/compose/install/) if you don't have it yet.
-3. Run `docker-compose up --build -d` in the root directory.
+3. Run `./couchdb/couchdb-init.sh` in the root directory.
 
 This should launch a new CouchDB instance on `http://localhost:5984`, create system database, configure CouchDB as Single Node, enable CORS, create `hospitalrun` database, create a default admin with a username of `admin` and password of 'password', create a sample user with a username of `username` and password of 'password' to use new login page [#2137](https://github.com/HospitalRun/hospitalrun-frontend/pull/2137).
 
 Go to `http://localhost:5984/_utils` in your browser to view Fauxton and perform administrative tasks.
 
 **_Cleanup_**
-To delete the development database, go to the root of the project and run `docker-compose down -v --rmi all --remove-orphans`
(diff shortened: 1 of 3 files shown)


.github/CONTRIBUTING.md
```diff
@@ -18,14 +18,14 @@ root of this project to launch CouchDB. Below are the steps:
 
 1. Start [Docker](https://docs.docker.com/get-docker/). Install it if you don't have it yet.
 2. Install [Docker Compose](https://docs.docker.com/compose/install/) if you don't have it yet.
-3. Run `docker-compose up --build -d` in the root directory.
+3. Run `./couchdb/couchdb-init.sh` in the root directory.
```
fox1t, line 21: We need to add that this works only on Unix based systems. On Windows, we still need to let users type `docker-compose up --build -d` directely.

tobireuen: You are right, I will add this.

tehKapa (commented):
@tobireuen thank you for the PR! 
However, I don't understand this feature (in particular, the difference between `docker-compose up --build -d` command and `sh ./couchdb/couchdb-init.sh` command). Doesn't add anything, doesn't create a real shortcut, can only be done on Unix-like OS and hides the real command executed.

tobireuen:
> @tobireuen thank you for the PR!
> However, I don't understand this feature (in particular, the difference between `docker-compose up --build -d` command and `sh ./couchdb/couchdb-init.sh` command). Doesn't add anything, doesn't create a real shortcut, can only be done on Unix-like OS and hides the real command executed.

I found it to be a more convenient way to init and cleanup the database.
I could add cmd files for Windows users as well if that's a problem.

If you still think these changes are not useful, feel free to reject the PR and I'll keep the scripts for my personal use.

fox1t:
IMO it lowers the barrier where possibile. It would be nice to have a .bat for Windows :/

fox1t:
@tobireuen are you able to add scripts for windows too? If so it would be a nice addition to this PR. Otherwise we will merge it without and fix it in the future.

fox1t (approved):
LGTM

tobireuen:
> @tobireuen are you able to add scripts for windows too? If so it would be a nice addition to this PR. Otherwise we will merge it without and fix it in the future.

I will add the Windows scripts later today or tomorrow. So please wait for the merge and thank you for your feedback.

Issue #2341, opened 2020-08-27
Refactor patient appointments to use react query instead of redux


jackcmeyer:
## 🚀 Feature Proposal
Refactor patient appointments to use react query instead of redux

Need one hook: 
- usePatientAppointments (given a patient id, returns a list of the patient appointments)

## Motivation
Remove redux

## Technical Notes
- use `usePatientDiagnoses` as inspiration
- remove redux code after

Tomastomaslol:
Hi @jackcmeyer. I would love to take a look at this. Could you assign this to me?

jackcmeyer:
@Tomastomaslol you got it!

PR #2349, opened 2020-08-29
refactor(patient): refactor patient appointments to use react query

Tomastomaslol:
Fixes: #2341 

- add new hook `usePatientAppointments` and supporting tests
- Update `AppointmentsList` to use `usePatientAppointments` hook and update supporting tests
- Remove no longer used `fetchPatientAppointments`

### DIFF
--- a/src/__tests__/patients/appointments/AppointmentsList.test.tsx
+++ b/src/__tests__/patients/appointments/AppointmentsList.test.tsx
@@ -1,6 +1,6 @@
 import * as components from '@hospitalrun/components'
 import { Table } from '@hospitalrun/components'
-import { mount } from 'enzyme'
+import { mount, ReactWrapper } from 'enzyme'
 import { createMemoryHistory } from 'history'
 import React from 'react'
 import { act } from 'react-dom/test-utils'
@@ -64,13 +64,13 @@ const setup = async (patient = expectedPatient, appointments = expectedAppointme
 
   wrapper.update()
 
-  return wrapper
+  return { wrapper: wrapper as ReactWrapper }
 }
 
 describe('AppointmentsList', () => {
   describe('Table', () => {
     it('should render a list of appointments', async () => {
-      const wrapper = await setup()
+      const { wrapper } = await setup()
 
       const table = wrapper.find(Table)
 
@@ -100,7 +100,7 @@ describe('AppointmentsList', () => {
     })
 
     it('should navigate to appointment profile on appointment click', async () => {
-      const wrapper = await setup()
+      const { wrapper } = await setup()
       const tr = wrapper.find('tr').at(1)
 
       act(() => {
@@ -114,7 +114,7 @@ describe('AppointmentsList', () => {
 
   describe('Empty list', () => {
     it('should render a warning message if there are no appointments', async () => {
-      const wrapper = await setup(expectedPatient, [])
+      const { wrapper } = await setup(expectedPatient, [])
       const alert = wrapper.find(components.Alert)
 
       expect(alert).toHaveLength(1)
@@ -125,7 +125,7 @@ describe('AppointmentsList', () => {
 
   describe('New appointment button', () => {
     it('should render a new appointment button', async () => {
-      const wrapper = await setup()
+      const { wrapper } = await setup()
 
       const addNewAppointmentButton = wrapper.find(components.Button).at(0)
       expect(addNewAppointmentButton).toHaveLength(1)
@@ -133,7 +133,7 @@ describe('AppointmentsList', () => {
     })
 
     it('should navigate to new appointment page', async () => {
-      const wrapper = await setup()
+      const { wrapper } = await setup()
 
       await act(async () => {
         await wrapper.find(components.Button).at(0).simulate('click')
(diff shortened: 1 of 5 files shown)


src/__tests__/patients/appointments/AppointmentsList.test.tsx
```diff
@@ -45,26 +45,35 @@ const history = createMemoryHistory()
...
+        <Provider store={store}>
+          <AppointmentsList patientId={patient.id} />
+        </Provider>
+      </Router>,
+    )
+  })
+
+  wrapper.update()
+
   return wrapper
```
jackcmeyer, line 67: small refactor: let's change this to be `return { wrapper: wrapper as ReactWrapper }`. This makes the code more extensible in the future and gives type support for wrapper.

src/__tests__/patients/hooks/usePatientAppointments.test.tsx
```diff
@@ -0,0 +1,39 @@
+import usePatientAppointments from '../../../patients/hooks/usePatientAppointments'
+import PatientRepository from '../../../shared/db/PatientRepository'
+import Appointment from '../../shared/model/Appointment'
+import executeQuery from '../../test-utils/use-query.util'
+
+describe('use patient', () => {
```
jackcmeyer, line 6: change describe to `use patient appointments`

src/__tests__/patients/hooks/usePatientAppointments.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+  it('should return the patient', async () => {
```
jackcmeyer, line 7: test should be named `should return the patient's appointments

src/patients/appointments/AppointmentsList.tsx
```diff
@@ -51,10 +46,10 @@ const AppointmentsList = (props: Props) => {
       <br />
       <div className="row">
         <div className="col-md-12">
-          {appointments ? (
-            appointments.length > 0 ? (
+          {data ? (
+            data.length > 0 ? (
               <Table
-                data={appointments}
+                data={data}
```
jackcmeyer, line 52: seems like this can be simplified: 

```
if(data === undefined || status === 'loading') { return <Loading />}

if(data.length === 0) { // return warning message) }

// render the component
```

blestab (changes requested):
When a user is on a specific patient, 

image


and user clicks on New Appointment, 

image


The system doesn't remember what patient was currently selected and requires the user to re-select the patient. I think 

1) either the system should default the patient the user was on in the add appointment screen 
2) or when in the Appointments tab of the patient, clicking New Appointment should launch a modal. This way after appointment creation the user stays on the patient they were on in case they want to document something else on the same patient.

image

Tomastomaslol:
@jackcmeyer Thanks for your feedback much appreciated! I have implemented all your suggested improvements please let me know what you think when you got a spare minute!

Tomastomaslol:
@blestab 

sounds like there are opportunities here to improve the UX. I like both suggestions but I'm not sure what would be the best way forward. Maybe we could deal with this in a separate issue? I think there is some value in keeping PRs single purposed and as small as possible to make auditing and reviewing quicker and easier.

blestab:
> @blestab
> 
> sounds like there are opportunities here to improve the UX. I like both suggestions but I'm not sure what would be the best way forward. Maybe we could deal with this in a separate issue? I think there is some value in keeping PRs single purposed and as small as possible to make auditing and reviewing quicker and easier.

Agree 100%, in that case let's keep the current workflow and just default in the patient the user was on when we open the New Appointment screen

blestab (changes requested):
It looks like you removed the New Appointment button? Let's put it back, we will log a separate issue for the fixing of the bug.

image

src/patients/appointments/AppointmentsList.tsx
```diff
@@ -1,37 +1,46 @@
...
     },
   ]
+
   useAddBreadcrumbs(breadcrumbs)
 
-  useEffect(() => {
-    dispatch(fetchPatientAppointments(patientId))
-  }, [dispatch, patientId])
+  if (data === undefined || status === 'loading') {
+    return <Spinner color="blue" loading size={[10, 25]} type="ScaleLoader" />
```
jackcmeyer, line 32: We've got a `Loading` component in this project that keeps loading spinners consistent for this case.

jackcmeyer:
@blestab @Tomastomaslol I have created #2350 for an enhancement to "remember" the patient

blestab (dismissed):

Tomastomaslol:
> It looks like you removed the New Appointment button? Let's put it back, we will log a separate issue for the fixing of the bug.

@blestab thank you for pointing this out! I think I got too excited while refactoring. I added a test to make sure the mistake can't be repeated again.

https://github.com/HospitalRun/hospitalrun-frontend/pull/2349/files#diff-acbb980a7cb7a9dc63dd7576b9a802f9R127

src/patients/appointments/AppointmentsList.tsx
Tomastomaslol: I updated this my latest commit. please let me know what think of it

PR #2550, opened 2021-01-18
feat(visits): add "request imaging button" that auto-populate imaging…

Fibii:
… request

- Remove "add visit" button from visits tab (since there's one already) and add a "add imaging request" button instead
- Implement logic that populates the fields in `NewImagingRequest` if the user is redirected there from the visits tab

### DIFF
--- a/src/patients/visits/VisitTab.tsx
+++ b/src/patients/visits/VisitTab.tsx
@@ -1,11 +1,6 @@
-import { Button } from '@hospitalrun/components'
 import React, { useState } from 'react'
-import { useSelector } from 'react-redux'
 import { Route, Switch } from 'react-router-dom'
 
-import useTranslator from '../../shared/hooks/useTranslator'
-import Permissions from '../../shared/model/Permissions'
-import { RootState } from '../../shared/store'
 import AddVisitModal from './AddVisitModal'
 import ViewVisit from './ViewVisit'
 import VisitTable from './VisitTable'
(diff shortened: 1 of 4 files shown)


src/patients/visits/VisitTab.tsx
```diff
@@ -15,26 +10,9 @@ interface Props {
 }
 
 const VisitTab = ({ patientId }: Props) => {
-  const { t } = useTranslator()
-  const { permissions } = useSelector((state: RootState) => state.user)
   const [showAddVisitModal, setShowAddVisitModal] = useState(false)
   return (
     <>
-      <div className="row">
-        <div className="col-md-12 d-flex justify-content-end">
```
blestab, line 24: >
> Remove "add visit" button from visits tab (since there's one already) and add a "add imaging request" button instead
>
Thank @Fibii 
kindly note that this was not in the issue specification. This button is needed for UI/UX consistency with the rest of the design.

blestab (changes requested):

blestab:
>
> There should be a new button added to the patient visits page which allows a user to make a new imaging request quickly.
>
I cannot yet see where this was implemented on the visits page. I am guessing since a new imaging request is done against a patient's visit, I am guessing this would be a button on each visit line item under the Actions column. I'm checking with the UX/UI team on where the button should be placed
cc @lauggh @StefanoMiC @jackcmeyer

PR #2552, opened 2021-01-20
fix(lab): In New lab request make the visit field required

### LINKED ISSUE #2551: Lab - In the Request Lab routine make the visit field required
## 🐛 Bug Report

When **Request Lab** for a patient without any visit, it will crash.

## To Reproduce

1. Create an patient without adding a visit
2. Click **Request Lab**, input the patient field, leave the visit field blank since it is not required.
3. Notice the crash/error (see screenshot in comments)

## Expected behavior

Request Lab should behave similar to **New Imaging Request** i.e. make the visit field required

## Technical Details

Make the visit field required (as in Imaging Request)

## Your Environment

- *node version*: 12
- *os*: Windows

SamuelQZQ:
Fixes #2551 .

- I make the `NewLabRequest` use the exact same method to solve the `null` visit issue. [`NewImagingRequest.tsx` line65](https://github.com/HospitalRun/hospitalrun-frontend/blob/45647d3b4a15ca8d9c57ebda66b3712658d62f19/src/imagings/requests/NewImagingRequest.tsx#L65)
- My concern is, we use the `as Option[]` at line 57. So that the type of the variable `visits` is not right. It should be `Option[] | null`. Maybe we should do some refactor in `NewImagingRequest.tsx` and `NewLabRequest.tsx` to fix this wrong type. Otherwise, it may lead to other bugs in the future.

### DIFF
--- a/src/__tests__/labs/requests/NewLabRequest.test.tsx
+++ b/src/__tests__/labs/requests/NewLabRequest.test.tsx
@@ -122,7 +122,7 @@ describe('New Lab Request', () => {
       expect(selectInput).toBeInTheDocument()
       expect(selectInput).toHaveDisplayValue([''])
       expect(selectLabel).toBeInTheDocument()
-      expect(selectLabel).not.toHaveAttribute('title', 'This is a required input')
+      expect(selectLabel).toHaveAttribute('title', 'This is a required input')
     })
 
     it('should render a save button', () => {
--- a/src/labs/requests/NewLabRequest.tsx
+++ b/src/labs/requests/NewLabRequest.tsx
@@ -51,7 +51,7 @@ const NewLabRequest = () => {
 
   const onPatientChange = (patient: Patient) => {
     if (patient) {
-      const visits = patient.visits.map((v) => ({
+      const visits = patient.visits?.map((v) => ({
         label: `${v.type} at ${format(new Date(v.startDateTime), 'yyyy-MM-dd hh:mm a')}`,
         value: v.id,
       })) as Option[]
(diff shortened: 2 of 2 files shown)


tehKapa:
Could you please sign our CLA? Thank you https://cla.js.foundation/HospitalRun/hospitalrun-frontend?pullRequest=2552

SamuelQZQ:
I have signed the CLA

blestab (changes requested):
>
>I have opened a new session in another chrome instance and I can see the visit option, which is not required, which i guess if why this bug occurs. So if you make the visit required then the issue should be resolved.



Looking good @SamuelQZQ I can see we are no longer crashing if there is no visit specified, however since the visit is required for each lab request we need to make the visit field required as it is on the imaging request.

image

image

blestab (commented):
Thanks @SamuelQZQ 

This is now behaving the same as the Request Imaging. I did notice however that Request Imaging (like this latest change) does not prevent saving when visit is not selected, even though it shows visit as required. 

Will create separate issues for these though unless you want to deal with the Lab issue in this PR?

image

chadian:
This somewhat relates to my PR #2555 to make sure `visits` is an empty array when a patient is created. This would help with the "crashing" part but doesn't change the fact if it should be a required field or not (what is covered in this PR).

SamuelQZQ:
@blestab I prefer to work on it in a separate issue. Let's complete this one. I will test more carefully in future work.

blestab:
> @blestab I prefer to work on it in a separate issue. Let's complete this one. I will test more carefully in future work.

Awesome! thanks @SamuelQZQ 

@tehKapa this one now looks good.

blestab (approved):

blestab:
> This somewhat relates to my PR #2555 to make sure `visits` is an empty array when a patient is created. This would help with the "crashing" part but doesn't change the fact if it should be a required field or not (what is covered in this PR).

Yeah just noticed that. I have since created a separate issue for enforcing the 'required visit' behavior

PR #2554, opened 2021-01-26
fix(medication): set correct quantity and value for new medications

### LINKED ISSUE #2492: Medication: New medication request quantity amount and unit not creating correctly
## 🐛 Bug Report

When a new medication request is created, the quantity (value and unit) do not seem to be created correctly. 
This leads to the edit/view form not displaying the quantity value and unit. 
Editing however works as expected.

## To Reproduce

- Access staging: https://staging.hospitalrun.io/
- Create a new medication request, specifying a quantity amount and unit. NOTE: You may have to create a patient first if none exists.
- Go to the medication requests listing
- View the request you just created and notice the Quantity just displays "x" instead of amount x unit e.g. "15 x TAB"

image

image

## Expected behavior

- New medication request should correctly update the quantity amount and unit when saving such the the display and edit screen can also access these values e.g. editing the above example to Quantity amount 30, unit CAPSULE, correctly updates the model
- See how edit request correctly updates these and fix the request medication functionality to do the same thing.

image

image

## Your Environment

- *node version*: 6,8,10
- *fastify version*: >=1.0.0
- *os*: Windows
- *any other relevant information*

chadian:
Fixes #2492.

Fixes setting the quantity and value properties on the medication model via the form `NewMedicationRequest`.

- Fix the way quantity and value is set on a new medication model from the form


blestab (changes requested):
Code changes LGTM as they appear to resolve the issue. May you fix the lint errors in src/medications/requests/NewMedicationRequest.tsx andwe should be all set.

chadian:
@blestab All set! I had to fix the extends for the eslint rules for `eslint-config-prettier` to work in disabling conflicting rules (in this case `jsx-curly-newline`)

src/scheduling/appointments/AppointmentDetailForm.tsx
```diff
@@ -134,10 +133,8 @@ const AppointmentDetailForm = (props: Props) => {
               label={t('scheduling.appointment.reason')}
               value={appointment.reason}
               isEditable={isEditable}
-              onChange={
-                (event: React.ChangeEvent<HTMLTextAreaElement>) =>
-                  onFieldChange && onFieldChange('reason', event.currentTarget.value)
-                // eslint-disable-next-line react/jsx-curly-newline
```
chadian, line 140: This ignore is no longer needed since it was conflicting with `prettier/react`. With prettierrc extends updated `eslint-config-prettier` is able to disable the conflicting prettier rule, opting for `jsx-curly-newline` in this case.

blestab (dismissed):
LGTM

src/scheduling/appointments/AppointmentDetailForm.tsx
tehKapa: Ok, remove it!

jackcmeyer (changes requested):
I would have expected a test to be modified/added to show that this is working

src/__tests__/medications/requests/NewMedicationRequest.test.tsx
```diff
@@ -184,20 +194,45 @@ describe('New Medication Request', () => {
...
+
+      userEvent.type(quantityUnit, '200')
+      userEvent.type(quantityValue, 'mg')
 
       userEvent.click(
         screen.getByRole('button', {
           name: /medications\.requests\.new/i,
         }),
       )
-      expect(history.location.pathname).toEqual('/medications/new')
```
chadian, line 200: I think this test was previously a false positive. The test description `should save the medication request and navigate to "/medications/:id"` but the submission was failing and the URL never progressed to the `/medications/:id`. I believe the test is now fixed and checks that on the submitted page that quantity unit and quantity value are present.

chadian:
@jackcmeyer Added a test to go along with these changes

tehKapa:
@chadian  nIeed someone to review it

PR #2565, opened 2021-02-10
fix(viewpatient): keep tabs active for nested routes

### LINKED ISSUE #2563: Focus of menu item "Visits" disappears when I view the details
## To Reproduce

When I view a visit detail tab, the focus of the "Visits" menu item disappears preventing me from understanding in which section I'm in.

Right behavior

Screenshot 2021-02-09 at 12 38 19

Wrong behavior

Screenshot 2021-02-09 at 12 38 25

agusbrand:
Fixes #2563 

**Note:** The bug was initially reported for the patient **visits tab**, but I realized that the same bug was also happening in the tabs: **allergies, notes, care-plans, care-goals**. Fixed all of them.

**Why the bug happened?**
When someone visits one of the mentioned tabs and, for example, clicks an element to view it, and it opens in a nested route, then the tab loses the active status because the route path changed.

- fix: Keep corresponding tabs as active for nested routes
- Add missing warning alert in visits table (attached screenshots)

Warning alert before (shows empty table):
image

Warning alert after:
image

### DIFF
--- a/src/__tests__/patients/view/ViewPatient.test.tsx
+++ b/src/__tests__/patients/view/ViewPatient.test.tsx
@@ -293,4 +293,20 @@ describe('ViewPatient', () => {
       expect(screen.getByText(/patient\.careGoals\.warning\.noCareGoals/i)).toBeInTheDocument()
     })
   })
+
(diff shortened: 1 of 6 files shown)


tehKapa (approved):
LGTM

PR #2660, opened 2021-04-23
feat(translate blood-type to ru)

la83lynx:
feat(translate blood-type to ru)


tehKapa:
Hi @la83lynx. 
Could you create a single PR with all the Russian translations? Another thing: remember to use semantic commit for the commit message.
Example: 

`chore(i18n): adds Russian lang for bloodtype`

PR #2673, opened 2021-05-12
Update Spanish Translation

verygreentea:
New Spanish translation, old one was incomplete.

Before approving this PR, make sure this merges correctly. I was unable to get the locales files through Git, so there may be issues when merging.

- Updating the Spanish translation, which was incomplete. I wasn't sure of the UI context of all strings, but I will update the translation if and when necessary. 

n/a, did not modify code.


src/locales/es/translations/actions/index.ts
```diff
@@ -0,0 +1,15 @@
...
+    edit: 'Editar',
+    save: 'Guardar',
+    update: 'Actualizar',
+    complete: 'Finalizar',
+    delete: 'Borrar',
+    cancel: 'Cancelar',
+    new: 'Nuevo',
+    list: 'Enumerar',
+    search: 'Buscar',
+    confirmDelete: 'Borrar Confirmación',
```
cpondoc, line 13: Not sure how many translations you want to add, but looking strictly at the English translations, it appears as though you are missing a couple within `actions/index.ts`. Specifically, look at:

- next
- previous
- page
- add
- view
- logout

src/locales/es/translations/index.ts
```diff
@@ -0,0 +1,19 @@
+import actions from './actions'
+import dashboard from './dashboard'
+import patient from './patient'
+import patients from './patients'
+import scheduling from './scheduling'
+import states from './states'
+import sex from './sex'
+import labs from './labs'
+
+export default {
```
cpondoc, line 10: Similar to earlier issue, but if we are talking strictly in terms of being comprehensive, I believe you are missing a couple of translation folders. Specifically, terms such as `bloodType` and `medications` are missing. I would also refer to the full English translations here: https://github.com/HospitalRun/hospitalrun-frontend/blob/master/src/shared/locales/enUs/translations/index.ts.

src/locales/es/translations/labs/index.ts
```diff
@@ -0,0 +1,30 @@
+export default {
+  labs: {
+    label: 'Pruebas de laboratorio',
```
cpondoc, line 3: I may leave a comment on a couple of these to link you to the English translation files in `shared/locales` to point you to any incompleted fields :)
https://github.com/HospitalRun/hospitalrun-frontend/blob/master/src/shared/locales/enUs/translations/labs/index.ts

cpondoc (changes requested):
Everything looks good on my end! Two things:

- Not sure if you have to re-clone the repository or simply do something like a `git pull`, but the files should be located within `src/shared/locales/es.` Most of the files actually already exist, although you may need to create some new files in order to complete all of the translations and stay up to date with the most recent version of the front-end.
- I would just look again at all of the translations, if you get the time. All in all, though, these look **great**!

cpondoc:
@morrme Any other thoughts on this PR?

PR #2689, opened 2021-06-02
fix(2494): Incidents: Add the ability to edit incident details

### LINKED ISSUE #2494: Incidents: Add incident edit form
## 🚀 Feature Proposal

Add the ability to edit a reported incident.

## Motivation

One an incident has been reported, the system must provide a way to update it, to correct any mistakes or provide a progress update.

#### Incident Model

Add the following new fields to the incident model

```
reportedTo: string
```

#### Incident View Form

Make the following fields editable on the current view form: department, category, category item, description, incident reported to, patient impacted (introduced via #2493)

##### Footer 
- Add a new footer button "Update" which will save changes made by the user.

riiniii:
fix #2494

Fixes #2494.

- merge code from master to current branch
- change useResolveIncident hook to useUpdateIncident
- in ViewIncidentDetails page, implement controlled inputs/forms with useState and useEffect, and make relevant fields editable: 
        1a. department
	1b. category
	1c. category item
	1d. description
	1e. incident reported to
	1f. patient impacted
- make ViewIncidentDetails.test.tsx reflect that form should be editable
- make it so that patient field is always visible, with on view/update, allows user to update the patient field from previously null/empty/undeff to a new patient value
- add reportedTo


riiniii:
cla was signed but showing as not signed, waiting on this

riiniii:
@tehKapa could you add james + cynthia again for the reviewers? I can't seem to add them by myself

DrewGregory:
@riiniii The issue with the CLA is that every author every commit in this PR has to sign the CLA. I would edit the commits to remove the authorship of Renee Huang, as it seems that they do not have a GitHub account and thus cannot sign the CLA. @tehKapa It seems like a good amount of contributors are running into this -- should the CLA only consider the GitHub author of the PR?

src/__tests__/incidents/hooks/useUpdateIncident.test.tsx
```diff
@@ -1,13 +1,12 @@
 import subDays from 'date-fns/subDays'
 
-import useResolveIncident from '../../../incidents/hooks/useResolveIncident'
+import useUpdateIncident from '../../../incidents/hooks/useUpdateIncident'
```
DrewGregory, line 3: I like that you made editing and resolving under the same mutation abstraction.

src/incidents/hooks/useUpdateIncident.tsx
```diff
@@ -3,15 +3,23 @@ import { queryCache, useMutation } from 'react-query'
 import IncidentRepository from '../../shared/db/IncidentRepository'
 import Incident from '../../shared/model/Incident'
 
+function getResolvedOn(incident: Incident) {
```
DrewGregory, line 6: This works, but makes it a little bit harder to type check. You might want to add some return types here so that it is obvious what state is resolved versus not resolved.

DrewGregory (dismissed):
Looks good. I would add a return type to `getResolvedOn()` so that it is clear to a user of this function what an "unresolved" result would look like.

src/incidents/view/ViewIncidentDetails.tsx
```diff
@@ -22,27 +24,42 @@ function ViewIncidentDetails(props: Props) {
...
+        <Button className="mr-2" color="success" onClick={onUpdate} key="incidents.reports.update">
+          {t('incidents.reports.update')}
+        </Button>,
+        <Button
+          onClick={onResolve}
+          className="mr-2"
+          color="primary"
+          key="incidents.reports.resolve"
+        >
           {t('incidents.reports.resolve')}
```
cynthiachen7, line 63: To clarify, is the button here supposed to say resolve or update? (in the translations, since you added an "update" field)

src/labs/ViewLab.tsx
```diff
@@ -129,7 +129,7 @@ const ViewLab = () => {
 
     if (permissions.includes(Permissions.CompleteLab)) {
       buttons.push(
-        <Button className="mr-2" onClick={onComplete} color="primary" key="labs.requests.complete">
+        <Button className="mr-2" color="primary" onClick={onComplete} key="labs.requests.complete">
```
cynthiachen7, line 132: Any particular reason for switching order here?

src/incidents/view/ViewIncidentDetails.tsx
```diff
@@ -125,22 +176,36 @@ function ViewIncidentDetails(props: Props) {
           <TextFieldWithLabelFormGroup
             label={t('incidents.reports.description')}
             name="description"
-            value={data.description}
+            value={incident.description}
+            isEditable
+            onChange={onIncidentChange('description')}
```
cynthiachen7, line 181: Clean!

cynthiachen7 (dismissed):

src/incidents/view/ViewIncidentDetails.tsx
riiniii: ah, let me double check!

src/labs/ViewLab.tsx
riiniii: not particularly, just to mtch the order of the buttons.push a few lines above

riiniii:
updated w/ pr fixes/comment updates, not in progress anymore

src/incidents/view/ViewIncidentDetails.tsx
riiniii: whoops, i left both buttons. one for "update" which lets user edit most fields, and resolve which just updates the status to resolved.

PR #2785, opened 2021-10-20
WIP: fix(patients): allow the duplicate patient modal to show properly on …

### LINKED ISSUE #2546: Duplicate Patient Modal cannot be triggered
## 🐛 Bug Report

While converting NewPatient.test.tsx to RTL, we found that the "Duplicate Patient Modal" is impossible to trigger by a user. It looks like it performs a de-dupe against the redux store, but there is nothing that ever populates that store with patients? Thus de-dupe never finds a duplicate.

Creating based on comment in https://github.com/HospitalRun/hospitalrun-frontend/pull/2516#issuecomment-753378004

## To Reproduce

Uncomment the "Duplicate Patient Modal" test and run it.

## Expected behavior

The test would pass.

connordmccandless:
…possible duplicate patient

fix #2546

- duplicatePatientList in DuplicateNewPatientModal.tsx properly updated to an array of patients as opposed to a single patient
- NewPatient.tsx now uses the usePatients hook instead of the state (that isn't being updated) to properly retrieve the current list of all patients


xmile1 (approved):

tehKapa:
@connordmccandless is still in wip?

connordmccandless:
@tehKapa I left a comment on the ticket when I last worked on the issue. I fixed the issue described in the ticket, but it seems that the written tests are not written correctly and will always return a failure. I wasn't sure if rewriting the test code would be part of the ticket.

